PR #96, opened 2017-02-01
adds skeleton to enhance documentation

cc10512:
Adds configuration to support doxygen documentation generation to
autotools as well as an initial skeleton to integrate documentation
that is not part of the source code.

For now, I kept it simple so that we can just add .md files as
"chapters" for different aspects of the documentation. Add a chapter
by creating a new file, prefixed by the chapter number (e.g. Chapter 1
Overview is 01_overview.md). The chapter number prefix is used instead
of creating a hierarchy of doxygen pages because those introduce extra
links that are a pain. However, doxygen consumes markdown very nicely,
and the only need is to place the files in the appropriate order.

If we want to be fancy with the documents generated (PDF and HTML)
then we need more work on the style sheets and we'll likely required
to generate custom header and footer files -- this will be a task for a
technical writer.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,4 @@
 # DOXYGEN SUPPORT
-# include $(top_srcdir)/aminclude.am
 @DX_RULES@
 
 ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS} -I m4
@@ -24,7 +23,8 @@ AM_DISTCHECK_CONFIGURE_FLAGS = --with-bmv2 --with-fe-cpp
 
 # ensure the distribution of the doxygen configuration file
 EXTRA_DIST = $(top_srcdir)/docs/doxygen.cfg \
-	     $(builddir)/doxygen-out/html $(builddir)/doxygen-out/pi.pdf
+	     $(top_srcdir)/docs/bfn_latex.sty \
+	     $(top_srcdir)/docs/*.md
 
 mostlyclean-local:
 	-$(RM) -r $(DX_CLEANFILES)
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,7 @@
+# DOXYGEN SUPPORT
+# include $(top_srcdir)/aminclude.am
+@DX_RULES@
+
 ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS} -I m4
 
 if WITH_CLI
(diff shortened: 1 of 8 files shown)


Makefile.am
```diff
@@ -17,3 +21,10 @@ DIST_SUBDIRS = third_party lib include src generators targets CLI examples \
 bin tests frontends_extra $(MAYBE_PROTO)
 
 AM_DISTCHECK_CONFIGURE_FLAGS = --with-bmv2 --with-fe-cpp
+
+# ensure the distribution of the doxygen configuration file
+EXTRA_DIST = $(top_srcdir)/docs/doxygen.cfg \
+	     $(builddir)/doxygen-out/html $(builddir)/doxygen-out/pi.pdf
```
antoninbas: what about `docs/bfn_latex.sty` and the md files, shouldn't they be included (if someone wants to rebuild the doc)?
you can't distribute the built files however, otherwise `make dist` fails. I suppose you could add a manual `dist` dependency on the doc, but I wouldn't do that. Instead we should have the Travis job build the documentation and upload it.

Makefile.am
```diff
@@ -17,3 +21,10 @@ DIST_SUBDIRS = third_party lib include src generators targets CLI examples \
...
+
+mostlyclean-local:
+	-$(RM) -r $(DX_CLEANFILES)
```
antoninbas: is this a reasonable candidate for `mostlyclean`, why not `CLEANFILES = $(DX_CLEANFILES)`

Makefile.am
```diff
@@ -1,3 +1,7 @@
+# DOXYGEN SUPPORT
+# include $(top_srcdir)/aminclude.am
```
cc10512: left over from an earlier version of the ax_prog_doxygen macro that was splitting the features and the makefile rules. will remove.

Makefile.am
cc10512: yes, the distribution should include all sources. a binary distribution should include the built documentation, and it is better to have it automatically built.

Makefile.am
cc10512: adding it to CLEANFILES or MOSTLYCLEANFILES does not remove directories, so it actually does not clean. it has only the -f flag, not the -rf.

antoninbas: got it

Makefile.am
antoninbas: `make dist` is for a source distribution though

cc10512:
@antoninbas pushed the changes you requested.

PR #100, opened 2017-02-08
P4runtime table read

antoninbas:
Add support for P4Runtime's TableRead

This required adding support for parsing table entries in the C++ PI frontend, which is achieved by MatchKeyReader and ActionDataReader. The DeviceMgr was updated to leverage these and now implements table_read and table_read_all.

Unit tests were added to test entry retrieval. I took advantage of this commit to cleanup the match table unit tests (less code duplication by using gtest value-parametrized tests).


antoninbas:
@teverman I don't know if this is of much interest to you. It simply implements DeviceMgr::table_read, which you can now use.

PR #101, opened 2017-02-09
Update p4runtime.proto

wmohsin:
Add RPCs for configuration of a P4 target.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -15,7 +15,7 @@
 syntax = "proto3";
 
 import "google/rpc/status.proto";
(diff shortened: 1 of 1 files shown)


proto/p4/p4runtime.proto
```diff
@@ -15,6 +15,7 @@
 syntax = "proto3";
 
 import "google/rpc/status.proto";
+import "config/p4info.proto";
```
antoninbas: pretty sure this needs to be "p4/config/p4info.proto"

wmohsin: Done.

PR #102, opened 2017-02-10
Add PI support for fetching action profile entries

antoninbas:


antoninbas:
Will add corresponding support to `DeviceMgr` in a subsequent PR

PR #109, opened 2017-02-27
P4info update

antoninbas:
Each commit in this pull request implements one of the following:
- Generalize MatchField definition in p4info
- Remove HeaderField and HeaderFieldList from p4info
- Use Preamble in Param message definition
- Introduce path in Preamble message in p4info
- Add with_entry_timeout to Table message in p4info
- Add has_const_default_action_entry to p4info

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -84,11 +84,8 @@ message Table {
 message Action {
   Preamble preamble = 1;
   message Param {
-    // id is not really needed, as it can be inferred from the action id and the
-    // position of the parameter, but having it could be convenient.
-    int32 id = 1;
-    string name = 2;
-    int32 bitwidth = 3;
+    Preamble preamble = 1;
+    int32 bitwidth = 2;
   }
   repeated Param params = 2;
 }
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -40,7 +40,7 @@ message Preamble {
   // somewhere. This also means that ids of different types can be mixed
   // (e.g. direct resource list for a table) without ambiguity. Note that id 0
   // is reserved and means "invalid id".
-  int32 id = 1;
+  uint32 id = 1;
   // fully qualified name of the P4 object, e.g. c1.c2.ipv4_lpm
   string path = 2;
   // an alias for the P4 object, probably shorter than its path. The only
@@ -75,9 +75,9 @@ message Table {
   repeated MatchField match_fields = 2;
   // even when the table is indirect (see implementation_id) below, this field
   // includes all possible actions for the table
-  repeated int32 action_ids = 3;
+  repeated uint32 action_ids = 3;
   // 0 (default value) means that the table does not have a const default action
-  int32 const_default_action_id = 4;
+  uint32 const_default_action_id = 4;
   // a table may have a const default action, whose action parameter values can
   // be changed at runtime. However, in most cases a table's action entry is set
   // in the P4 program and cannot be changed, which is what this boolean flag
@@ -87,11 +87,11 @@ message Table {
   // (default value) means that the table is a regular (direct) match table. As
   // of today, only action profiles are supported but other table
   // implementations may be added in the future
-  int32 implementation_id = 6;
+  uint32 implementation_id = 6;
   // ids of the direct resources (if any) attached to this table; for now this
   // includes only direct counters and direct meters, but other resources may be
   // added in the future
-  repeated int32 direct_resource_ids = 7;
+  repeated uint32 direct_resource_ids = 7;
   int64 size = 8;  // number of entries in table
   bool with_entry_timeout = 9;  // entry ageing is enabled for table
 }
@@ -109,7 +109,7 @@ message ActionProfile {
   Preamble preamble = 1;
   // the ids of the tables sharing this action profile; note that action profile
   // sharing is not yet supported by the P4 software switch (bmv2)
-  repeated int32 table_ids = 2;
+  repeated uint32 table_ids = 2;
   // true iff the action profile used dynamic selection
   bool with_selector = 3;
   int64 size = 4;  // number of entries in action profile
@@ -119,7 +119,7 @@ message Counter {
   Preamble preamble = 1;
   // 0 (default value) if the counter is not direct, the id of the table to
   // which the counter is attached otherwise
-  int32 direct_table_id = 2;
+  uint32 direct_table_id = 2;
   // in the current P4 spec the attribute is called 'type', I believe 'unit' is
   // more appropriate for this and I hope to see the spec updated
   enum Unit {
@@ -138,7 +138,7 @@ message Meter {
   Preamble preamble = 1;
   // 0 (default value) if the meter is not direct, the id of the table to which
   // the meter is attached otherwise
-  int32 direct_table_id = 2;
+  uint32 direct_table_id = 2;
   // in the current P4 spec the attribute is called 'type', I believe 'unit' is
   // more appropriate for this and I hope to see the spec updated
   enum Unit {
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -79,10 +79,10 @@ message Table {
   // 0 (default value) means that the table does not have a const default action
   uint32 const_default_action_id = 4;
   // a table may have a const default action, whose action parameter values can
-  // be changed at runtime. However, in most cases a table's action entry is set
-  // in the P4 program and cannot be changed, which is what this boolean flag
-  // indicates.
-  bool has_const_default_action_entry = 5;
+  // be changed at runtime. However, in most cases the parameters of the default
+  // action are also bound at compile-time and cannot be changed by the runtime,
+  // which is what this boolean flag indicates.
+  bool const_default_action_has_mutable_params = 5;
   // P4 id of the "implementation" for this table (e.g. action profile id); 0
   // (default value) means that the table is a regular (direct) match table. As
   // of today, only action profiles are supported but other table
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -21,13 +21,11 @@ import "google/protobuf/any.proto";
 package p4.config;
 
 message P4Info {
-  repeated HeaderField header_fields = 1;
-  repeated HeaderFieldList header_field_lists = 2;
-  repeated Table tables = 3;
-  repeated Action actions = 4;
-  repeated ActionProfile action_profiles = 5;
-  repeated Counter counters = 6;
-  repeated Meter meters = 7;
+  repeated Table tables = 1;
+  repeated Action actions = 2;
+  repeated ActionProfile action_profiles = 3;
+  repeated Counter counters = 4;
+  repeated Meter meters = 5;
 
   // for extern?
   repeated google.protobuf.Any externs = 100;
@@ -46,16 +44,6 @@ message Preamble {
   // fully qualified name of the P4 object
   string name = 2;
   repeated string annotations = 3;
-}
-
-message HeaderField {
-  Preamble preamble = 1;
-  int32 bitwidth = 2;
-}
-
-message HeaderFieldList {
-  Preamble preamble = 1;
-  repeated int32 header_field_ids = 2;
 }
 
 // TODO(antonin): define inside the Table message?
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -60,7 +60,8 @@ message HeaderFieldList {
 
 // TODO(antonin): define inside the Table message?
 message MatchField {
-  int32 header_field_id = 1;
+  Preamble preamble = 1;
+  int32 bitwidth = 2;
   enum MatchType {
     UNSPECIFIED = 0;
     VALID = 1;
@@ -69,7 +70,7 @@ message MatchField {
     TERNARY = 4;
     RANGE = 5;
   }
-  MatchType match_type = 2;
+  MatchType match_type = 3;
 }
 
 message Table {
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -41,9 +41,18 @@ message Preamble {
   // (e.g. direct resource list for a table) without ambiguity. Note that id 0
   // is reserved and means "invalid id".
   int32 id = 1;
-  // fully qualified name of the P4 object
-  string name = 2;
-  repeated string annotations = 3;
+  // fully qualified name of the P4 object, e.g. c1.c2.ipv4_lpm
+  string path = 2;
+  // an alias for the P4 object, probably shorter than its path. The only
+  // constraint is for it to be unique with respect to other P4 objects of the
+  // same type. By default, the compiler uses the shortest suffix of the path
+  // that uniquely indentifies the object. For example if the P4 program
+  // contains two tables with paths s.c1.t and s.c2.t, the default names will
+  // respectively be c1.t and c2.t. The P4 programmer may also override the
+  // default name for any P4 object (TBD). When resolving a P4 object id, an
+  // application should be able to indiscriminately use the path or the name.
+  string name = 3;
+  repeated string annotations = 4;
 }
 
 // TODO(antonin): define inside the Table message?
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -88,6 +88,7 @@ message Table {
   // added in the future
   repeated int32 direct_resource_ids = 6;
   int64 size = 7;  // number of entries in table
+  bool with_entry_timeout = 8;  // entry ageing is enabled for table
 }
 
 message Action {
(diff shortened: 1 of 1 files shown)


antoninbas:
Obviously Travis regressions are failing given that I changes the p4info.proto format. Once an agreement is reached on the update, I will fix all the dependent code and update the pull request.

proto/p4/config/p4info.proto
```diff
@@ -43,24 +41,24 @@ message Preamble {
   // (e.g. direct resource list for a table) without ambiguity. Note that id 0
   // is reserved and means "invalid id".
   int32 id = 1;
```
wmohsin: Suggest s/int32/uint32/ for ID type to be consistent with P4Runtime.

proto/p4/config/p4info.proto
```diff
@@ -43,24 +41,24 @@ message Preamble {
...
+  // application should be able to indiscriminately use the path or the name.
+  string name = 3;
+  repeated string annotations = 4;
 }
 
 // TODO(antonin): define inside the Table message?
 message MatchField {
-  int32 header_field_id = 1;
+  Preamble preamble = 1;
+  int32 bitwidth = 2;
```
wmohsin: Please add a short comment to clarify meaning/semantics.

proto/p4/config/p4info.proto
```diff
@@ -80,26 +78,29 @@ message Table {
   repeated int32 action_ids = 3;
```
wmohsin: suggest s/int32/uint32/ for ID type (see my comment above)

proto/p4/config/p4info.proto
```diff
@@ -80,26 +78,29 @@ message Table {
...
   // 0 (default value) means that the table does not have a const default action
   int32 const_default_action_id = 4;
```
wmohsin: suggest s/int32/uint32/ for ID type (see my comment above)
[Here and elsewhere]

proto/p4/config/p4info.proto
```diff
@@ -43,24 +41,24 @@ message Preamble {
...
-  int32 bitwidth = 2;
-}
-
-message HeaderFieldList {
-  Preamble preamble = 1;
-  repeated int32 header_field_ids = 2;
+  // fully qualified name of the P4 object, e.g. c1.c2.ipv4_lpm
+  string path = 2;
+  // an alias for the P4 object, probably shorter than its path. The only
+  // constraint is for it to be unique with respect to other P4 objects of the
```
wmohsin: "The only constraint is for it to be unique with respect to other P4 objects of the same type"

Is it just "same type"? Seems like that would disallow c1.c2.PuntTable and c1.c3.PuntTable ... which is not what we want.
Should it be worded "of the same type at the same level"?

proto/p4/config/p4info.proto
```diff
@@ -80,26 +78,29 @@ message Table {
...
-  int32 implementation_id = 5;
+  int32 implementation_id = 6;
   // ids of the direct resources (if any) attached to this table; for now this
   // includes only direct counters and direct meters, but other resources may be
   // added in the future
-  repeated int32 direct_resource_ids = 6;
-  int64 size = 7;  // number of entries in table
+  repeated int32 direct_resource_ids = 7;
+  int64 size = 8;  // number of entries in table
+  bool with_entry_timeout = 9;  // entry ageing is enabled for table
```
wmohsin: "Entry Aging" seem new (at least to me). Is this coming from P4_16? Seems like we need to discuss implications on P4Runtime (I would imagine controller specifying the age).

proto/p4/config/p4info.proto
```diff
@@ -80,26 +78,29 @@ message Table {
...
 
 message Action {
   Preamble preamble = 1;
   message Param {
-    // id is not really needed, as it can be inferred from the action id and the
-    // position of the parameter, but having it could be convenient.
-    int32 id = 1;
-    string name = 2;
-    int32 bitwidth = 3;
+    Preamble preamble = 1;
```
wmohsin: If we are using Preamble for action parameters, does the notion of path extend to them? I would imagine no (since function args in C++ are not namespace-typed), and so we should probably change the comments for Preamble.name above to not call it "an alias for the P4 object" but its given name at the local (or innermost namespace/control) level.
What do you think?

proto/p4/config/p4info.proto
```diff
@@ -80,26 +78,29 @@ message Table {
...
+    int32 bitwidth = 2;
```
wmohsin: What does this mean if I'm passing "struct" args?

proto/p4/config/p4info.proto
```diff
@@ -21,13 +21,11 @@ import "google/protobuf/any.proto";
 package p4.config;
 
 message P4Info {
-  repeated HeaderField header_fields = 1;
```
wmohsin: With these two gone, how does a fixed-function target tie the table match-fields to what they semantically mean. For example:
header ipv4_base_t  foo;
reads "foo.ttl"
How (based solely on P4Info) does a target deduce that foo.ttl actually means an IPv4 TTL field?
(I believe we we thinking that we would standardize Header/HeaderField names and keep a separate canonical mapping from those to target-specific "qualifiers" of what these fields stand for)

proto/p4/config/p4info.proto
sethfowler: That seems like a specific instance of a more general problem. We have no concept of type for action parameters. If we introduced that concept, presumably we could also specify how to serialize a parameter of a given type. At the moment, though, parameters are basically just sequences of bits.

proto/p4/config/p4info.proto
```diff
@@ -41,9 +41,18 @@ message Preamble {
...
+  string path = 2;
+  // an alias for the P4 object, probably shorter than its path. The only
+  // constraint is for it to be unique with respect to other P4 objects of the
+  // same type. By default, the compiler uses the shortest suffix of the path
+  // that uniquely indentifies the object. For example if the P4 program
+  // contains two tables with paths s.c1.t and s.c2.t, the default names will
+  // respectively be c1.t and c2.t. The P4 programmer may also override the
+  // default name for any P4 object (TBD). When resolving a P4 object id, an
+  // application should be able to indiscriminately use the path or the name.
+  string name = 3;
```
sethfowler: Looks like we now have three unique identifiers for every object - `path`, `name`, and `id` must all be unique. `path` and `name`, per this comment, are also in the same namespace (otherwise they couldn't be used indiscriminately). This seems like a lot to provide at this layer, but I *can* see the use of offering multiple names to describe the same object. Thinking from a language design perspective, if we're going to go this route, I'd much rather get rid of the concept of a "canonical" name entirely and just replaced both `path` and `name` with `repeated string names`. This will allow us to give the programmer a lot more flexibility when exposing names to the control plane.

proto/p4/config/p4info.proto
```diff
@@ -78,17 +78,22 @@ message Table {
   repeated int32 action_ids = 3;
   // 0 (default value) means that the table does not have a const default action
   int32 const_default_action_id = 4;
+  // a table may have a const default action, whose action parameter values can
+  // be changed at runtime. However, in most cases a table's action entry is set
+  // in the P4 program and cannot be changed, which is what this boolean flag
+  // indicates.
+  bool has_const_default_action_entry = 5;
```
sethfowler: This comment and field name is a bit unclear to me even though I know why you're adding this. =) I feel like our terminology is inconsistent (action params / action data / action entry), which may be the problem. Maybe rename to something more verbose but clearer like `const_default_action_has_mutable_params` and update the comment to state more clearly that the issue is whether it's possible to change the parameters to the const default action.

This is so awkward that I wonder if it's even the right approach. Maybe when the action parameters cannot be changed, we should just wrap the original action in a synthesized an action that has no parameters.

proto/p4/config/p4info.proto
sethfowler: In both cases you ultimately just have a string, right? You'll always need to have semantic knowledge of the P4 program (or at least, knowledge of some convention that the P4 program adheres to) to interpret them. Seems like whether we regard those strings as "header fields" or "match keys" doesn't change much in the end. (If I'm missing something, though, I'd love to know more!)

proto/p4/config/p4info.proto
sethfowler: I don't know much about it myself, but I do have the general comment that there's an alternative to support this feature: we can treat it as an extern, and just reference it in `direct_resource_ids`. We may want to be cautious about baking more stuff into `Table` than is necessary.

proto/p4/config/p4info.proto
antoninbas: I updated all the IDs as you suggested.

proto/p4/config/p4info.proto
antoninbas: @wmohsin I believed we discussed this today during our talk. This mapping would be the responsibility of the p4c backend for the target. We can discuss this more on Friday or by email with @teverman if there are more concerns.

proto/p4/config/p4info.proto
antoninbas: That would disallow using `PuntTable` twice for 2 different tables. Bu the names could be `c2.PuntTable` and `c1.c3.PuntTable`, or `PuntTable1` and `PuntTable2` (if overridden by programmer).
My understanding is that for your use case, you would be using the `path` anyway (and organize namespaces accordingly) and would not use `name`.

proto/p4/config/p4info.proto
antoninbas: @wmohsin entry ageing is supported in P4_14 (it is a standard table property called `support_timeout`). It enables ageing for this table's entries. If ageing is enabled, p4runtime.proto would be able to provide a TTL for entries when doing a `TableWrite` (TTL can vary between entries).

@sethfowler this could belong in a separate map of properties, but not really in `direct_resource_ids`. We may want to be cautious about producing a runtime representation of the program which is too generic to be convenient to use.

sethfowler: The common case should be easy, that's true.

proto/p4/config/p4info.proto
antoninbas: I am not convinced that the notion of path really extends to action params and match fields (but all the other fields of `Preamble` do apply). We don't have to use `Preamble` here, I can revert to the old code.
Apart from the action param case, it seems that we are in disagreement over path vs. name. As I mentioned above, from our discussions I believe that you don't really have a need for the `name` (you basically just need the `path`). For you the `name` should just be the last component of the `path`. However I do have use cases in mind for this `name` alias. In particular I do believe the ability for the P4 programmer to choose an alias for a P4 object that will be independent of the P4 hierarchy and resilient to code-refactoring is important.

proto/p4/config/p4info.proto
antoninbas: I like your proposed name so I'll update the pull request.

Regarding the second part of your comment, I believe this is a **P4** problem that has not been solved: the ability to rename actions in the scope of a table when binding parameters at compile-time.

proto/p4/config/p4info.proto
antoninbas: IMO the notion of `path` is important because it preserves the P4 hierarchy in p4info. A use case we are considering is generating some C++ code based on the p4info (sort of like a better PD). This path can be use to define namespaces in the generated code in a meaningful way.

Another use case, which in my opinion is orthogonal, is the ability for the P4 programmer to choose a runtime name for a P4 object. This name could be use by a runtime application and in the event of some simple code-refactoring (e.g. renaming a control block) would mean that the application doesn't need to be updated.

So the way I see it, `path` and `name` are both valuable and cannot be bundled together. I almost replaced `name` with a repeated field of names as you are suggesting when designing the pull request, and I am still open to this suggestion, but I don't know if it would play well with the use case I describe above.

proto/p4/config/p4info.proto
antoninbas: What do you have in mind?

proto/p4/config/p4info.proto
teverman: Yes, Antonin is correct - we should be able to do this in the p4c backend for the switch.  As we discussed on Friday, the p4c IR still contains all the header field data, it just never gets into P4Info.

proto/p4/config/p4info.proto
teverman: Regarding the preamble data in a MatchField - if this matches a header field, then the id would be the ID of the header field, as assigned by p4c or annotation, and then name would be the name of the field from the header_type definition, correct?

proto/p4/config/p4info.proto
teverman: In order to make this work without header fields in the P4Info, we will need to store the HF IDs in our proto blob for the switch.  Where in the p4c IR does the authoritative ID exist, i.e. the one that would appear in the P4Info header field if one was present?  Is this ID present before the serializer runs, or does it appear after.  Does the serializer make ID decisions such that we could get different ID output depending on whether the command-line chooses to invoke the serializer?

sethfowler: The ID does not really exist in the IR at all. It's purely a function of the P4 object's name and type. The serializer doesn't make ID decisions; it just follows an algorithm for generating IDs.

(There is one exception - `@id` annotations are, of course, present in the IR.)

proto/p4/config/p4info.proto
sethfowler: We should definitely clarify what the intent is here. One tricky thing is that it's possibly to use an `@name` annotation to name a match field, and that name might be different from the name of the corresponding header field.

In general, my expectation (this is my personal view) is that by default, the name of the match key would match the field name, but I would expect the IDs for match keys to live in a different "ID namespace".

antoninbas: @teverman if we go through with this update, there won't be a need for header fields to be assigned an id any more. My plan is to have match fields receive an id much like action parameters do (or some other scheme if desired). The reasoning is that if header fields 1) are not referenced by any other object in p4info and 2) don't need to be referenced by p4runtime, then there is no need for them to even appear in p4info and no need for them to be assigned an id. Do you see an issue with that approach?

proto/p4/config/p4info.proto
antoninbas: @teverman I think I covered this in my answer to another comment. There won't be p4info ids for header fields any more if we go through with this update because header fields will not play any part in p4runtime

teverman:
When we get a table write, it has an ID for each match key.  We need data
to tell us how to interpret the match key and transform it into a form
that's suitable for the vendor SDK.  We expect to lookup that data based on
the table match key ID(s).  The output of the lookup is the data generated
by the p4c backend that was passed to the switch at config time.

The open issue is how we establish the linkage between the runtime match
key ID and the data that the p4c backend generates.  The backend should be
able to produce a lot of this data by inspection of parser and header
objects in the IR.

antoninbas:
@teverman You will still be able to get this information:
1. you will analyze header declarations and parser code to generate a mapping header field (name) to fixed protocol field
2. you will then inspect table match keys in the IR. When a match field is an expression which simply evaluates a header field, you will use the mapping in step 1 to find which fixed protocol field is matched on in this table. You can finally use p4info.proto to generate the final mapping <match field id> -> <fixed protocol field>. If I am not mistaken, your backend has access to the match field ids because each backend is responsible for invoking the API that generates p4info.

teverman:
OK, so let me see if I have this right Antonin.

For your #1, the key to the generated data, let's call it a Field Desciptor
(FD), has to be the header field name.  The IR does not have any usable
integer ID that is also present in the P4Info.

Per #2:
- Our backend always produces P4Info to use internally, regardless of
whether the user wants to save it to a file.  (We are doing this now
anyway).
- Given a P4Info table match key, the backend needs to extract the name
from the match preamble and use it to find the FD by name.  We then have
two choices:
1) The backend can add the P4Info ID to the FD, giving the switch
information to directly lookup the FD by P4 runtime match key ID, or
2) At runtime, the switch can do a match key ID lookup in the P4Info,
extract the match key's name from the matching preamble, and do a name
lookup on the FD.

antoninbas:
@teverman yes that's exactly what I have in mind. For the second step you mentioned 2 alternatives, with me describing essentially the first one in my comment. I didn't want to make things confusing by giving 2 alternatives so I stuck with the first one :). But you can totally compute that mapping at the switch level and not include it "as is" in your backend output.

teverman:
With all of this settled, I'd like to suggest a multi-phase approach to
introducing this change:

1) Update P4Info to contain a preamble in the match key.  Leave full header
field support in place for the time being.  Have p4c generate P4Info with
the match key preambles. Commit to github.
2) Pull phase 1 changes to partner repos and adapt P4Info client code to
operate with the new match key preamble, and then phase out dependencies on
P4Info header fields.  Commit these changes to partner internal repose.
3) Remove header fields from P4Info, which should now no longer be
referenced. Commit to github.
4) Pull phase 3 changes and confirm partner code is free of header field
dependencies.

antoninbas:
@teverman So in step 1), `MatchField` would keep using the same id as before, but we would have 2 copies of it: the existing `header_field_id` field for backward-compatibility and the `id` field in the `Preamble`.
Then in step 2), we can remove the notion of header field id and the tight coupling of match key fields with header fields?
I can definitely open a separate pull request, identical to this one except for the `HeaderField` removal.
Did I understand you correctly?

teverman:
Yes, and in step 1, to make sure we cab close the loop, the match key's
preamble name also needs to be populated to whatever you expect it to be in
step 2.

antoninbas:
Replaced by #111

PR #112, opened 2017-03-06
Update p4runtime.proto

wmohsin:
1. Batching of write/read RPCs for various P4 entities
2. New RPC/semantics for P4 config

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -72,6 +72,12 @@ service P4Runtime {
   rpc ActionProfileRead(ActionProfileReadRequest)
       returns (stream ActionProfileReadResponse) {
   }
+  // Modifies 'CounterEntry's.
+  rpc CounterWrite(CounterWriteRequest) returns (stream CounterWriteResponse) {
+  }
+  // Reads 'CounterEntry's.
+  rpc CounterRead(CounterReadRequest) returns (stream CounterReadResponse) {
+  }
 }
 
 //------------------------------------------------------------------------------
@@ -266,6 +272,9 @@ message CounterEntry {
     TableEntry table_entry = 3;
   }
   CounterData data = 4;
+
+  // TODO(wmohsin) Remove once partners have moved to new API.
+  repeated CounterCell cells = 5;
 }
 
 message CounterData {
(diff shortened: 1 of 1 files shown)


antoninbas:
@wmohsin the Travis build is broken because of a syntax error in your `p4runtime.proto`

antoninbas:
@wmohsin any chance of leaving the old counter RPCs for now? I think doing so would mean current code would keep working, and that would make the transition easier for me?

wmohsin:
OK, I can leave the counter RPCs in. I removed them based on the feedback from our Fri meeting. However, I see it is causing Travis to break.

antoninbas:
In my case, it is only the meter RPC that I'm not using currently. But I think @aghaffarkhah is not using either one, which probably explains the confusion. Sorry about that.

wmohsin:
Travis seems green now.

proto/p4/p4runtime.proto
```diff
@@ -50,88 +78,56 @@ service P4Runtime {
...
+  Uint128 election_id = 2;
+  repeated Update updates = 3;
 }
 
-message P4ConfigGetResponse {
-  config.P4Info p4info = 1;
-  // Target-specific P4 configuration.
-  bytes p4_device_config = 2;
+message WriteResponse {
 }
```
antoninbas: we are not longer returning a repeated field of `google.rpc.Status`?
what's the best practice for returning an error status now?

proto/p4/p4runtime.proto
```diff
@@ -191,157 +187,50 @@ message Action {
...
-  // Read all meters if empty.
-  // Read all cells of meter.meter_id if meter.cells is empty.
-  // 'MeterEntry' is shared between request and response messages. For
-  // the request, meters.cells.config need not be populated, and is ignored.
-  repeated MeterEntry meters = 2;
-}
-
+// For WriteRequest, Update.Type has following semantics:
+//  - INSERT: Insert the cell in the meter array (with given configuration)
+//  - MODIFY: Modify the meter cell configuration
```
antoninbas: I still have trouble understanding the semantics of `INSERT` vs `MODIFY` for counters and meters. It seems that in my case at least, the implementation would be exactly the same.
But I guess that applies to tables as well. Would it make sense to have a single idempotent write operation that would replace `INSERT` and `MODIFY`? I think it may make sense from the controller point-of-view but may make switch implementation more difficult...

proto/p4/p4runtime.proto
```diff
@@ -50,88 +78,56 @@ service P4Runtime {
   // Reads 'CounterEntry's.
   rpc CounterRead(CounterReadRequest) returns (stream CounterReadResponse) {
   }
```
aghaffarkhah: FYI I do not use the Counter Read/Write message. If it is OK with Antonin you can remove.

antoninbas: @aghaffarkhah no I'm using them, we can remove them at the same time as the rest when we cleanup

antoninbas:
I'm merging this pull request. Based on today's meeting, I believe that we will take care of error reporting in a subsequent pull request.

PR #115, opened 2017-03-11
Don't use Preamble for match fields and params

antoninbas:
In p4info.proto. We reserve preamble for top-level P4 objects. The plan
is also to make match fields' and action parameters' ids local to the
table / action they belong to.


antoninbas:
@teverman are you ok with this change we discussed by email?

PR #118, opened 2017-03-15
PI changes after p4info.proto update

antoninbas:
- fields and field lists are no longer part of p4info (which means that learning is broken for the moment, as it used field lists)
- match fields' and action params' ids are now locally scoped to tables / actions
- support for alias


antoninbas:
compiler support has been added (https://github.com/p4lang/p4c/pull/369), so I'm merging this

PR #121, opened 2017-03-21
Remove HeaderField and HeaderFieldList from p4info

antoninbas:


antoninbas:
@sethfowler let's merge this first so you can point to the merged commit in p4c

PR #124, opened 2017-03-24
P4runtime update

antoninbas:
Update DeviceMgr and demo code to support the new p4runtime.proto RPCs.


antoninbas:
@teverman this updates the DeviceMgr API in reaction to the update to p4runtime.proto. This pull request does not remove the old APIs yet, as I'll wait for your feedback before doing so.

PR #126, opened 2017-03-31
Conditionally generate Python code for .proto

antoninbas:
We have updated the build system to conditionally generate Python
protobuf / gRPC bindings for p4runtime.proto, p4info.proto,... The
Python code will also be installed. We do not generate code if the
Python gRPC plugin cannot be found.


antoninbas:
@teverman any objection to me merging this? We use Python for P4 unit tests.

teverman:
Go ahead and merge it.  We've had to do some rework here to get the
grpc/proto cpp working in our environment.  I suspect Python will be
similar, so I should be able to deal with it.

PR #128, opened 2017-03-31
Proposal to support extern types in p4runtime

antoninbas:
Proposed changes to p4info.proto and p4runtime.proto. We leverage
google.protobuf.Any to pack / unpack extern-specific data in both
cases. Guidelines to choose packages / appropriate URLs for
extern-specific protobuf messages are TBD.

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -21,14 +21,12 @@ import "google/protobuf/any.proto";
 package p4.config;
 
 message P4Info {
-  repeated Table tables = 1;
-  repeated Action actions = 2;
-  repeated ActionProfile action_profiles = 3;
-  repeated Counter counters = 4;
-  repeated Meter meters = 5;
-
-  // for extern?
-  repeated google.protobuf.Any externs = 100;
+  repeated Extern externs = 1;
+  repeated Table tables = 2;
+  repeated Action actions = 3;
+  repeated ActionProfile action_profiles = 4;
+  repeated Counter counters = 5;
+  repeated Meter meters = 6;
 }
 
 message Preamble {
@@ -53,6 +51,23 @@ message Preamble {
   // application should be able to indiscriminately use the name or the alias.
   string alias = 3;
   repeated string annotations = 4;
+}
+
+// used to group all extern instances of the same type in one message
+message Extern {
+  // the extern_type_id is assigned during compilation. It is likely that this
+  // id will in fact come from a P4 annotation to the extern declaration and
+  // that each vendor will receive a prefix to avoid collisions.
+  uint32 extern_type_id = 1;
+  string extern_type_name = 2;
+  repeated ExternInstance instances = 3;
+}
+
+message ExternInstance {
+  Preamble preamble = 1;
+  // specific to the extern type, declared in a separate vendor-specific proto
+  // file
+  google.protobuf.Any info = 2;
 }
 
 // TODO(antonin): define inside the Table message?
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -14,6 +14,7 @@
 
 syntax = "proto3";
 
+import "google/protobuf/any.proto";
 import "google/rpc/status.proto";
 import "p4/config/p4info.proto";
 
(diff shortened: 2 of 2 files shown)


antoninbas:
There is no rush to review and merge this pull request. This simply implements the proposal I presented today. We still need to provide guidelines on URL choices for serialized messages.

proto/p4/config/p4info.proto
```diff
@@ -55,6 +53,22 @@ message Preamble {
   repeated string annotations = 4;
 }
 
+// used to group all extern instances of the same type in one message
+message Extern {
+  // the extern_type_id is assigned during compilation. It is likely that this
+  // id will in fact come from a P4 annotation to the extern declaration and
+  // that each vendor will receive a prefix to avoid collisions.
```
sethfowler: An alternative we could consider would be to generate the id as a hash of the name. The likelihood of a collision is pretty low when considering how many types of P4 externs are likely to be used within the same program.

proto/p4/config/p4info.proto
```diff
@@ -21,14 +21,12 @@ import "google/protobuf/any.proto";
...
 message P4Info {
-  repeated Table tables = 1;
-  repeated Action actions = 2;
-  repeated ActionProfile action_profiles = 3;
-  repeated Counter counters = 4;
-  repeated Meter meters = 5;
-
-  // for extern?
-  repeated google.protobuf.Any externs = 100;
+  repeated Extern externs = 1;
```
teverman: Why does externs come first, vs. being inserted after meters as field #6?

proto/p4/config/p4info.proto
```diff
@@ -55,6 +53,22 @@ message Preamble {
...
+  // that each vendor will receive a prefix to avoid collisions.
+  uint32 extern_type_id = 1;
+  string extern_type_name = 2;
+  repeated ExternInstance instances = 3;
+}
+
+message ExternInstance {
+  Preamble preamble = 1;
+  // specific to the extern type, declared in a separate proto file
+  google.protobuf.Any info = 2;
```
teverman: Would it be useful to add a comment clarifying that the separate proto file is likely to be vendor dependent?

proto/p4/p4runtime.proto
```diff
@@ -114,14 +115,21 @@ message Update {
 
 message Entity {
   oneof entity {
-    TableEntry table_entry = 1;
-    ActionProfileMember action_profile_member = 2;
-    ActionProfileGroup action_profile_group = 3;
-    MeterEntry meter_entry = 4;
-    CounterEntry counter_entry = 5;
+    ExternEntry extern_entry = 1;
```
teverman: Same as above - extern_entry first or after counter_entry?

proto/p4/config/p4info.proto
wmohsin: So this was discussed in our Friday weekly. The idea was to avoid "Extern" eventually getting buried in the middle as more fields are added (improved readability), since the bottom will grow and the top is fixed.

proto/p4/p4runtime.proto
wmohsin: Please see my comment above.

proto/p4/p4runtime.proto
```diff
@@ -114,14 +115,21 @@ message Update {
...
+    TableEntry table_entry = 2;
+    ActionProfileMember action_profile_member = 3;
+    ActionProfileGroup action_profile_group = 4;
+    MeterEntry meter_entry = 5;
+    CounterEntry counter_entry = 6;
   }
 }
 
+message ExternEntry {
+  uint32 extern_type_id = 1;
```
wmohsin: Suggest copying the comment from p4info.proto above to clarify this field.

proto/p4/config/p4info.proto
antoninbas: I believe this was @wmohsin 's suggestion. Since we are likely to add native support for standard externs (e.g. registers, qos?) to this list, it may make more sense to have externs at the beginning to be able to grow the list as much as we need to. That being said, we are unlikely to reach 100 standard objects. I don't have a strong preference.

proto/p4/config/p4info.proto
antoninbas: done

proto/p4/config/p4info.proto
antoninbas: I think we still have the issue of the id being too small for me to be comfortable with the idea. On a related note, do you know if there is a way in P4_16 to declare extern types in a specific namespace? 2 different vendors may wish to both declare a `counter` extern, and even though it is unlikely that they would both be used in the same P4 program, it would be nice to be able to have `vendor1.counter` and `vendor2.counter`.

proto/p4/p4runtime.proto
teverman: At some point, we are going to have to decide that P4Info and P4runtime have reached a level of maturity where renumbering fields is verboten.

wmohsin: Agree! That usually happens when this gets deployed and/or there are versioning concerns (old/new client/service, persisted proto dumps, etc. ... standard proto hygiene). I don't think we are there yet, but getting closer.

wmohsin: Agree! That usually happens when this gets deployed and/or there are versioning concerns (old/new client/service, persisted proto dumps, etc. ... standard proto hygiene). I don't think we are there yet, but getting closer.

proto/p4/config/p4info.proto
sethfowler: I don't expect that to be possible in P4-16 per se, but we are working on post-P4-16 extensions to the language that would allow that. I think we can be confident that a feature along those lines will be available.

It's true that we have only 32 bits for this field, which is too few to give high confidence that there will be no collisions. We could bump it up to 64 bits, I suppose, if we don't intend to transmit `extern_type_id` values very frequently in the API.

antoninbas:
@aghaffarkhah @wmohsin @teverman As discussed at our last meeting, I'd like to be able to merge this before we move on to other things (e.g. packet IO). This extern stuff is not set-in-stone, so we can revisit it later when we start actually using externs... If you have no objection, please approve the pull request.

PR #131, opened 2017-04-17
Factor out direct/indirect counters/meters into separate entities

wmohsin:
This was the PoR from our meeting/discussions, and consistent with P4_16 and v1model treatment of these externs.

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -26,7 +26,9 @@ message P4Info {
   repeated Action actions = 3;
   repeated ActionProfile action_profiles = 4;
   repeated Counter counters = 5;
-  repeated Meter meters = 6;
+  repeated DirectCounter direct_counters = 6;
+  repeated Meter meters = 7;
+  repeated DirectMeter direct_meters = 8;
 }
 
 message Preamble {
@@ -134,44 +136,57 @@ message ActionProfile {
   int64 size = 4;  // max number of entries in action profile
 }
 
-message Counter {
-  Preamble preamble = 1;
-  // 0 (default value) if the counter is not direct, the id of the table to
-  // which the counter is attached otherwise
-  uint32 direct_table_id = 2;
-  // in the current P4 spec the attribute is called 'type', I believe 'unit' is
-  // more appropriate for this and I hope to see the spec updated
+message CounterSpec {
+  // Corresponds to 'type' attribute for counter in P4 spec.
   enum Unit {
     UNSPECIFIED = 0;
     BYTES = 1;
     PACKETS = 2;
     BOTH = 3;  // not part of the P4 spec yet but will be in the future
   }
-  Unit unit = 3;
-  // number of entries in the counter array, for direct counters this is equal
-  // to the corresponding table's size
-  int64 size = 4;
+  Unit unit = 1;
 }
 
-message Meter {
+message Counter {
   Preamble preamble = 1;
-  // 0 (default value) if the meter is not direct, the id of the table to which
-  // the meter is attached otherwise
-  uint32 direct_table_id = 2;
-  // in the current P4 spec the attribute is called 'type', I believe 'unit' is
-  // more appropriate for this and I hope to see the spec updated
+  CounterSpec spec = 2;
+  // number of entries in the counter array
+  int64 size = 3;
+}
+
+message DirectCounter {
+  Preamble preamble = 1;
+  CounterSpec spec = 2;
+  // the id of the table to which the counter is attached
+  uint32 direct_table_id = 3;
+}
+
+message MeterSpec {
+  // Corresponds to 'type' attribute for meter in P4 spec.
   enum Unit {
     UNSPECIFIED = 0;
     BYTES = 1;
     PACKETS = 2;
   }
-  Unit unit = 3;
-  enum Type {  // not part of the P4 spec yet but will be in the future
+  // Not part of the P4 spec yet but will be in the future.
+  enum Type {
     COLOR_UNAWARE = 0;  // default value
     COLOR_AWARE = 1;
   }
-  Type type = 4;
-  // number of entries in the meter array, for direct meters this is equal to
-  // the corresponding table's size
-  int64 size = 5;
+  Unit unit = 1;
+  Type type = 2;
+}
+
+message Meter {
+  Preamble preamble = 1;
+  MeterSpec spec = 2;
+  // number of entries in the meter array
+  int64 size = 3;
+}
+
+message DirectMeter {
+  Preamble preamble = 1;
+  MeterSpec spec = 2;
+  // the id of the table to which the meter is attached
+  uint32 direct_table_id = 3;
 }
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -73,12 +73,6 @@ service P4Runtime {
   rpc ActionProfileRead(ActionProfileReadRequest)
       returns (stream ActionProfileReadResponse) {
   }
-  // Modifies 'CounterEntry's.
-  rpc CounterWrite(CounterWriteRequest) returns (stream CounterWriteResponse) {
-  }
-  // Reads 'CounterEntry's.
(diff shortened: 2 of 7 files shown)


proto/p4/config/p4info.proto
```diff
@@ -26,7 +26,9 @@ message P4Info {
   repeated Action actions = 3;
   repeated ActionProfile action_profiles = 4;
   repeated Counter counters = 5;
+  repeated DirectCounter direct_counters = 5;
```
antoninbas: you seem to be re-using the same proto ids here

proto/p4/config/p4info.proto
```diff
@@ -136,8 +138,22 @@ message ActionProfile {
 
 message Counter {
   Preamble preamble = 1;
-  // 0 (default value) if the counter is not direct, the id of the table to
-  // which the counter is attached otherwise
+  // in the current P4 spec the attribute is called 'type', I believe 'unit' is
+  // more appropriate for this and I hope to see the spec updated
+  enum Unit {
```
antoninbas: Let's move `Unit` outside of this message (and rename it `CounterUnit`) so it can be shared by direct & indirect?

proto/p4/config/p4info.proto
```diff
@@ -148,15 +164,30 @@ message Counter {
...
-  int64 size = 4;
 }
 
 message Meter {
   Preamble preamble = 1;
-  // 0 (default value) if the meter is not direct, the id of the table to which
-  // the meter is attached otherwise
+  // in the current P4 spec the attribute is called 'type', I believe 'unit' is
+  // more appropriate for this and I hope to see the spec updated
+  enum Unit {
```
antoninbas: Same as for counter unit

proto/p4/config/p4info.proto
```diff
@@ -148,15 +164,30 @@ message Counter {
...
+    UNSPECIFIED = 0;
+    BYTES = 1;
+    PACKETS = 2;
+  }
+  Unit unit = 2;
+  enum Type {  // not part of the P4 spec yet but will be in the future
```
antoninbas: Same as for unit

proto/p4/p4runtime.proto
```diff
@@ -135,8 +137,8 @@ message ExternEntry {
 
 // From Section 11 of P4_14 spec:
 // At run time, the table entry insert operation (not part of P4) must specify:
-// – Values for each field specified in the reads entry.
-// – The name of the action from the action_specification or the
+// â€“ Values for each field specified in the reads entry.
+// â€“ The name of the action from the action_specification or the
```
antoninbas: how did these characters get here?

proto/p4/config/p4info.proto
```diff
@@ -136,8 +138,22 @@ message ActionProfile {
 
 message Counter {
   Preamble preamble = 1;
-  // 0 (default value) if the counter is not direct, the id of the table to
-  // which the counter is attached otherwise
+  // in the current P4 spec the attribute is called 'type', I believe 'unit' is
+  // more appropriate for this and I hope to see the spec updated
```
teverman: This comment seems more like code review dialog than information to assist the developer.

proto/p4/config/p4info.proto
wmohsin: Done.

proto/p4/config/p4info.proto
wmohsin: Done.

proto/p4/config/p4info.proto
wmohsin: Done.

proto/p4/config/p4info.proto
wmohsin: Done.

proto/p4/config/p4info.proto
wmohsin: Done.

proto/p4/config/p4info.proto
```diff
@@ -134,44 +136,53 @@ message ActionProfile {
...
-    BYTES = 1;
-    PACKETS = 2;
-    BOTH = 3;  // not part of the P4 spec yet but will be in the future
-  }
-  Unit unit = 3;
-  // number of entries in the counter array, for direct counters this is equal
-  // to the corresponding table's size
+  CounterUnit unit = 2;
+  // number of entries in the counter array
   int64 size = 4;
```
antoninbas: there is no field with id 3

proto/p4/config/p4info.proto
```diff
@@ -134,44 +136,53 @@ message ActionProfile {
...
 }
 
+message DirectCounter {
+  Preamble preamble = 1;
+  // the id of the table to which the counter is attached
+  uint32 direct_table_id = 2;
+  CounterUnit unit = 3;
```
antoninbas: should we try to keep the same "layout" as for `Counter`: `Preamble -> unit -> [direct_table_id | size ]`?
same applies to meters below

proto/p4/p4runtime.proto
```diff
@@ -270,24 +281,31 @@ message MeterConfig {
...
-    // For direct counter, the associated table entry match.
-    // table_entry.action is ignored. Other fields specify the match.
-    TableEntry table_entry = 3;
-  }
-  CounterData data = 4;
-
+  int64 index = 2;
+  CounterData data = 3;
   // TODO(wmohsin) Remove once partners have moved to new API.
   repeated CounterCell cells = 5;
```
antoninbas: I think now may be a good time to remove `CounterCell`

proto/p4/config/p4info.proto
wmohsin: Done

proto/p4/p4runtime.proto
wmohsin: Done

proto/p4/config/p4info.proto
wmohsin: Makes sense

proto/p4/p4runtime.proto
antoninbas: I still see `CounterCell`, did you forget to push something?

wmohsin: Sorry, I'm learning the github code review UI :) I thought my comment wouldn't get published out until I PTAL at the end.
By the way, how about removing the Counter{Read|Write} RPCs as well (if you think CounterCell can be removed now)?

antoninbas:
@wmohsin I'm fine with removing Counter{Read|Write} (and same for meter) as part of this PR.

wmohsin:
Antonin: Re: enum conflicts (compiler failure), I would suggest we move them back as nested definitions. This potentially allows them to evolve differently (in line with our thinking of treating direct/indirect as separate top-level entities). I do not see significant maintenance overhead here.
The other option would be to encode entity specific prefix to disambiguate: COUNTER_UNIT_SPECIFIED and COUNTER_UNIT_BYTES vs METER_UNIT_UNSPECIFIED and METER_UNIT_BYTES ... but IMO this becomes repetitive and doesn't read as nicely.
One other option is to nest the enums in separate messages (so they are scoped and don't conflict), but this also seems unnecessary repetition between the message and enum name.
Let me know if you have any concerns.

antoninbas:
@wmohsin my personal favorite would be to encapsulate them in a separate message (one for counters, one for meters) to avoid the repetition (I honestly do not see how they could evolve differently in the future).
However I do not have a strong preference; but maybe others do?

wmohsin:
Yes, doing this now ...

wmohsin:
Friendly ping to other reviewers ...

teverman:
I see "All checks have failed".  Otherwise looks good.

antoninbas:
@teverman I will fix the regression checks one we have reached an agreement on the proto syntax

antoninbas:
I'm going to consider that everyone interested in reviewing this has given their review, and I will now cleanup the git history of this branch, and update the test code so that Travis regressions can pass. I will then merge the PR either later today or tomorrow. Please speak up if you have any objection.

PR #133, opened 2017-04-20
Reduce Docker image size

sethfowler:
This PR includes changes to reduce the size of the Docker image generated for this repo. The approach taken is pretty much the same as what we've done for other p4lang repos.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+**/.git/*
+**/.gitmodules
+**/.travis.yml
+**/Dockerfile
+**/*.md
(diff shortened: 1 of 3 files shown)


sethfowler:
These all fail with `clang-format` violations, but I didn't touch any source files. =\ Any idea what's happening there?

antoninbas:
these are not not clang-format violations, there are cpplint violations
the presence of the .git directory has an impact on cpplint (https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py#L117) and you are not including that directory in the docker image

sethfowler:
OK, I updated `.dockerignore` to ignore all the files under `.git`, but not `.git` itself. Locally, it looks like this fixed the problem.

PR #135, opened 2017-04-24
Add ActionRef message in p4info.proto

antoninbas:
Instead of having a repeated field of action ids in the Table message, we introduce a new message type, ActionRef, which encapsulates the action id with a list of annotations; this way each action reference can have its own annotations.


antoninbas:
@teverman I added a commit which restores the `action_ids` field (although with a different proto id). I'll make sure the compiler populates both. Will that work for you?

teverman:
Yes, that should work like we were able to do with the header field
transition.

antoninbas:
@teverman @samar-abdi The corresponding p4c pull request has been merged: https://github.com/p4lang/p4c/pull/521

PR #137, opened 2017-05-02
Config gnmi proto

antoninbas:
The first commit adds gnmi as a git dependency for this repository and adds the infra required to compile it.

The second is an attempt at an interface to process gnmi Get and Set requests. Nothing permanent, mostly a POC.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,6 @@
 	path = third_party/googletest
 	url = https://github.com/google/googletest
         ignore = untracked
+[submodule "proto/openconfig/reference"]
+	path = proto/openconfig/reference
+	url = https://github.com/openconfig/reference.git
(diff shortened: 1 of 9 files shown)


proto/frontend/PI/frontends/proto/config_mgr.h
```diff
@@ -0,0 +1,62 @@
...
+namespace pi {
+
+namespace fe {
+
+namespace proto {
+
+// forward declaration for PIMPL class
+class GnmiConfigMgrImp;
+
+class GnmiConfigMgr {
```
wmohsin: nit: The gNMI interface will be used for more than just Config. The other use-cases are (i) streaming telemetry (ii) streaming notifications to controller (iii) run-time control from controller.
Given that, perhaps call this GnmiMgr to not suggest that this is config-specific (unless you plan to create separate classes and plumbing for other use-cases).

gNMIServiceImpl gnmi_service

antoninbas: Thanks for the review, I changed the name as you suggested.

PR #139, opened 2017-05-10
Add support for writing meter configs in DeviceMgr

antoninbas:
The tricky case was direct meters, for which we need to be able to map
the match key to the correct entry handle; unless we want to start
adding PI C APIs which can work from the match key, like we did for
delete and modify (table entries). It doesn't seem like such a bad
solution to keep some limited per-entry state in DeviceMgr (may be
needed to support the controller_metadata cookie anyaway). To store this
state, the TableInfoStore was introduced.


antoninbas:
@teverman FYI I have started adding support for meters in DeviceMgr

PR #140, opened 2017-05-10
Fix p4runtime proto field numbering for MeterConfig

antoninbas:


antoninbas:
@teverman @wmohsin I don't think this will be a problem for anyone as I imagine you haven't started working on meters yet

PR #141, opened 2017-05-12
Make PI compilable in google3

chenxinming05:
This CL changes a place in proto/frontend/src/device_mgr.cpp to
 eliminate the "unused variable" error.

### DIFF
--- a/lib/vector.c
+++ b/lib/vector.c
@@ -98,9 +98,8 @@ void vector_remove_e(vector_t *v, void *e) {
 }
 
 void vector_destroy(vector_t *v) {
-  size_t index;
   if (v->clean_fn) {
-    for (index = 0; index < v->size; index++) {
+    for (size_t index = 0; index < v->size; index++) {
       v->clean_fn(access(v, index));
     }
   }
--- a/lib/vector.c
+++ b/lib/vector.c
@@ -98,8 +98,9 @@ void vector_remove_e(vector_t *v, void *e) {
 }
 
 void vector_destroy(vector_t *v) {
+  size_t index;
   if (v->clean_fn) {
-    for (size_t index = 0; index < v->size; index++) {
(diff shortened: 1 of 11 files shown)


sethfowler:
google3 is not compiling in C99 mode?

chenxinming05:
Not for pure c files. I can use an option -std=c99, but that breaks the strdup function.

third_party/p4lang_PI/src/p4info/actions.c:157:18: error: implicit declaration of function 'strdup' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  action->name = strdup(name);

So either we fix the for loop, or we fix the strdup function.

sethfowler:
It'd be much better to replace the call to `strdup` with code that uses `strncpy` or something similar. (Or figure out what #define is required to make `strdup` available with your compiler.)

antoninbas:
I am with @sethfowler on this. It would be much better to figure out how you can make strdup available. If this is a major roadblock, you can provide a strdup implementation under lib in this repo and use it instead; I believe that would be a better pull request. Declaring the loop iteration variable outside of the loop like this makes the code less clear.

ChrisDodd:
strdup should be defined in <string.h> according to POSIX (which has defined it pretty much from the start -- 1986?)

chenxinming05:
I did the trick by adding "-std=c11 -D_GNU_SOURCE" as the compiler parameter, so I'll discard the for loop changes.

strdup is not ANSI C so it's not visible in C99 and up by default.

chenxinming05:
Please take another look. I've updated the changes.

antoninbas:
I think that by default the compilers we use (locally or in regressions) all use `gnu11` as default, which is why we don't face the issue. The Google compiler probably conforms to ANSI C by default. I will open an issue for myself to document this requirement in the README.

PR #145, opened 2017-05-16
Controller metadata support

antoninbas:


antoninbas:
@teverman FYI I have added support for controller_metadata in DeviceMgr

PR #148, opened 2017-05-18
Tentative support for error_details in grpc::Status

antoninbas:
Since version 1.3.0, gRPC supports returning error details along with
the grpc::Status. This commit modifies the gRPC server implementation
for P4Runtime to leverage this feature. DeviceMgr methods return a
google::rpc::Status object which is converted to a grpc::Status object
using the grpc::SetErrorDetails helper function which ships with
gRPC. On the client side, the grpc::ExtractErrorDetails method can be
used to retrieve the error details.

Note that the google::rpc::Status proto contains a `repeated
google.protobuf.Any` field which will be used to carry
P4Runtime-specific error codes for each batch request, in addition to
the generic gRPC error code.

Because of the new dependency on error details, the configure.ac file
was updated to require a version >= 1.3.0 for gRPC.

### DIFF
--- a/proto/server/pi_server.cpp
+++ b/proto/server/pi_server.cpp
@@ -31,6 +31,7 @@
 #include <csignal>
 
 #include <grpc++/grpc++.h>
+#include <grpc++/support/error_details.h>
 
 #include "p4/p4runtime.grpc.pb.h"
 #include "google/rpc/code.pb.h"
@@ -69,6 +70,17 @@ StreamChannelClientMgr *packet_in_mgr;
 void packet_in_cb(DeviceMgr::device_id_t device_id, std::string packet,
                   void *cookie);
 
+
+// DeviceMgr::Status == google::rpc::Status
+grpc::Status to_grpc_status(const DeviceMgr::Status &from) {
+  grpc::Status to;
+  auto conversion_status = grpc::SetErrorDetails(from, &to);
+  // This can only fail if the second argument to SetErrorDetails is a nullptr,
+  // which cannot be the case here
+  assert(conversion_status.ok());
+  return to;
+}
+
 class P4RuntimeServiceImpl : public p4::P4Runtime::Service {
  private:
   Status Write(ServerContext *context,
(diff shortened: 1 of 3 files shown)


antoninbas:
@wmohsin @aghaffarkhah @teverman This is my proposal to update the "reference" gRPC server implementation to leverage the new error details support in gRPC.
This is based of my understanding on the new gRPC code. I have been looking at the tests in the github repo and I believe using `SetErrorDetails` is the right thing to do. However it means that the code needs to be linked with `libgrpc++_error_details`. I don't mind doing this; however if you see an issue I can just copy the code for this function to this repository. TBH I am a little confused by why this utility function is in its own library (https://github.com/grpc/grpc/blob/eceec8eaf6f38766364b203e92ccb6b25028b89b/BUILD#L167).

antoninbas:
@teverman Obviously this CL will wait until you have updated your grpc. I have been craving a way to be able to report my errors to the client :) We can discuss tomorrow and see if you have an ETA.

proto/server/pi_server.cpp
```diff
@@ -69,6 +70,17 @@ StreamChannelClientMgr *packet_in_mgr;
 void packet_in_cb(DeviceMgr::device_id_t device_id, std::string packet,
                   void *cookie);
 
+
+// DeviceMgr::Status == google::rpc::Status
+grpc::Status make_status(const DeviceMgr::Status &from) {
```
wmohsin: nit: Consider "to_grpc_status" or "grpc_status" to make it clear what 'status' type it is converting to.

antoninbas: done, thanks

antoninbas:
@samar-abdi did you have time to check if this can be merged in as discussed at our last meeting?

samar-abdi:
@antoninbas Sorry for the delay. Yes, it is ok to merge. We are on 1.3.0-dev on our third_party gRPC and grpc::SetErrorDetails / grpc::ExtractErrorDetails is implemented.

PR #154, opened 2017-05-30
Fix Range MatchField in p4runtime.proto

antoninbas:
We were using integers for the bounds, when it should be byte strings.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -142,6 +142,8 @@ message TableEntry {
 message FieldMatch {
   uint32 field_id = 1;
 
+  // Matches can be performed on arbitrarily-large inputs; the protobuf type
+  // 'bytes' is used to model arbitrarily-large values.
   message Exact {
     bytes value = 1;
   }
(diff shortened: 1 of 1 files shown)


proto/p4/p4runtime.proto
```diff
@@ -154,8 +154,8 @@ message FieldMatch {
     int32 prefix_len = 2;  // in bits
   }
   message Range {
```
wmohsin: Suggest a comment to clarify ... something like (taken from p4-16 spec):
A Range is logically a set that contains all values numerically between 'low' and 'high' inclusively.
P4_16 defines the two arguments ('low' and 'high') to be of type bit<W> or int<W>. Since W can be arbitrarily wide, this is modeled as protobuf type 'bytes'.

wmohsin: Hmm ... for some reason, angle brackets got eaten-up in my comment above.
I mean to say:
... to be of type bit{W} or int{W}
Replace curly-bracket and angle-bracket.

PR #156, opened 2017-06-05
Add sanity-checking for protobuf 'bytes' fields

antoninbas:
For match fields and action parameter values sent through P4Runtime, we
now check that:
 * the bytestring has the correct length (i.e. the number of bytes match
 the width of the P4 type)
 * for non byte-aligned field, the leading bits have been cleared
 (i.e. for a 12-bit field, the 4 leading bits of the first byte need to
 be cleared).

We also check that match keys have the appropriate number of match
fields and that they match the P4 table declaration. Same thing for
action parameters.

If one of these checks is violated, we return an INVALID_ARGUMENT
status.

These checks are now performed in DeviceMgr. Some of them may be
redundant with checks performed at lower levels of the implementation;
we can clean this up later if there is a performance concern.

### DIFF
--- a/proto/frontend/src/device_mgr.cpp
+++ b/proto/frontend/src/device_mgr.cpp
@@ -71,6 +71,23 @@ pi_meter_spec_t meter_spec_proto_to_pi(const p4::MeterConfig &config) {
   return pi_meter_spec;
 }
 
+uint8_t clz(uint8_t b) {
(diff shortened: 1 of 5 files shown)


antoninbas:
@teverman @wmohsin FYI, this is what we discussed at the last meeting

PR #161, opened 2017-06-19
Support ControllerPacketMetadata in DeviceMgr

antoninbas:
We added code to perform encap (packet-in) and decap (packet-out) for
controller packets. We support both cases: with and without metadata
(based on the p4info), with no overhead in the case without
metadata. This is a first attempt at controller metadata support, and
the code may be improved in the future for performance.

The DeviceMgr interface has been updated slightly to better match
p4runtime.proto.

This pull requests include unit tests to test the encap and decap
logic.


antoninbas:
@teverman FYI, since I have changed the DeviceMgr interface slightly, sorry for the large CL.

PR #182, opened 2017-07-25
fe_defines string sanitize

melshabani:


generators/fe_defines.c
```diff
@@ -93,9 +93,27 @@ static void to_upper(char *s) {
 }
 
 static void sanitize_name(char *s) {
+  char *valid_before = "$valid$";
+  char *valid_after = "valid";
+  char *valid_found = NULL;
+
+  while ((valid_found = strstr(s, valid_before)) != NULL) {
```
antoninbas: why don't you just use `memmove` to remove the `$` signs instead of making a copy?

melshabani: changed

generators/fe_defines.c
```diff
@@ -93,9 +93,23 @@ static void to_upper(char *s) {
...
+    // overwrite $valid$ with valid, leave out '/0'
+    memcpy(valid_found, valid_after, sizeof(valid_after) - 1);
+
+    // move over rest of string
+    char *src = &valid_found[sizeof(valid_before) - 1];
```
antoninbas: `sizeof` will return 8 here I think, which is the size of the pointer 
you need to change the definition of `valid_before` and `valid_after` to `char []`

melshabani: oops. you're right, fixed.

antoninbas: I meant that you should change
```
char *valid_before = "$valid$";
char *valid_after = "valid";
```
to
```
const char valid_before[] = "$valid$";
const char valid_after[] = "valid";
```
which would let you use `sizeof`
I didn't mean you should use variable-length arrays

PR #186, opened 2017-08-16
P4runtime direct resource support in TableEntry

antoninbas:


antoninbas:
I didn't go into too many details in the comments, I'm assuming we will cover all the edge cases in the spec. Let me know if you disagree with the text.

proto/p4/p4runtime.proto
```diff
@@ -136,6 +136,14 @@ message TableEntry {
   // this is stored, as long as it is returned with the rest of the entry in
   // a Read RPC.
   uint64 controller_metadata = 5;
+  // meter_config and counter_data are convenience fields that enable the
+  // controller to configure the direct resources associated with the table at
+  // the same time as a match-action entry is inserted or modified. When
+  // updating a table entry, leaving these fields unset means that the default
```
wmohsin: You probably meant s/default/direct/

antoninbas: Fixed

proto/p4/p4runtime.proto
```diff
@@ -136,6 +136,14 @@ message TableEntry {
   // this is stored, as long as it is returned with the rest of the entry in
   // a Read RPC.
   uint64 controller_metadata = 5;
+  // meter_config and counter_data are convenience fields that enable the
+  // controller to configure the direct resources associated with the table at
+  // the same time as a match-action entry is inserted or modified. When
+  // updating a table entry, leaving these fields unset means that the direct
+  // resources (if any) will not be updated. When reading a table entry, the
+  // values associated with the direct resources (if any) will be returned.
```
teverman: As I interpret this, I see a conflict between what this comment says and what Yavuz is expecting in the draft runtime spec.  The runtime spec says that the reply to a Read must equal the original Write request.  This comment states that the direct resources should be included in a Read, with no dependency on whether they were part of the original write.  If the controller chooses not to include them in the Write, then the reply violates the runtime spec if the controller adds any direct resource later.

proto/p4/p4runtime.proto
```diff
@@ -136,6 +136,14 @@ message TableEntry {
   // this is stored, as long as it is returned with the rest of the entry in
   // a Read RPC.
   uint64 controller_metadata = 5;
+  // meter_config and counter_data are convenience fields that enable the
+  // controller to configure the direct resources associated with the table at
+  // the same time as a match-action entry is inserted or modified. When
```
samar-abdi: Suggest adding:
If the table does not have a direct resource, then setting the direct resource field in any table write operation will return an error. If the table has a direct resource and a table entry is inserted without setting the direct resource field(s), the field(s) are assigned their default values.

proto/p4/p4runtime.proto
samar-abdi: I believe the intent is to return a value for the direct resource field in a table read if the direct resource is specified for the table. Added another comment for initializing the direct resource field(s), which should cover your concern.

proto/p4/p4runtime.proto
antoninbas: I agree with the first sentence. Not sure I understand the second sentence fully. Could you clarify what you mean by "their default values" (i.e. what are the default values for counters and meters)?

proto/p4/p4runtime.proto
antoninbas: We may need to clarify @yyetim text then. Let's say my table has a direct counter associated with it. If I add a table entry with a Write request (empty CounterData, so the counter's initial value is 0), then performs a Read on this entry, assuming some traffic went through and my flow entry was hit, then the counter value will not be 0 anymore, and obviously the message returned by Read is different from the message of the original Write.
@yyetim @wmohsin any thoughts on this?

proto/p4/p4runtime.proto
samar-abdi: I wanted to capture initialization. Rephrased:

"If the table has a direct resource and a table entry is inserted without setting the direct resource field, the field values are initialized to 0."

While this seems reasonable for counters, I am hoping for a better solution for direct meters. Perhaps, we should return an error if meter config is not set, similar to action parameters not being set? Accidentally specifying a peak rate of 0 to a meter would turn the meter red by default, which may end up black-holing the traffic if the action is to rate-limit on meter red!

proto/p4/p4runtime.proto
wmohsin: I agree, we should revise the "Write/Read symmetry" section to clarify this. I think the essence is (as I understand) that mutations done by a Write should be visible on a subsequent Read. That Read may additionally return additional data, which should be fine. So its not strict bit-by-bit symmetry but more like logical symmetry.

wmohsin: I agree, we should revise the "Write/Read symmetry" section to clarify this. I think the essence is (as I understand) that mutations done by a Write should be visible on a subsequent Read. That Read may additionally return additional data, which should be fine. So its not strict bit-by-bit symmetry but more like logical symmetry.

proto/p4/p4runtime.proto
antoninbas: I think it is up to the architecture (in this case PSA) to define default behavior for "uninitialized" resources. In the meter case, I seem to recall that the behavior would be to tag all packets as GREEN. Now I guess the question is what would be returned by the Read RPC in that case...

samar-abdi: Ok, we need a way to specify the resource's initial values when instantiating it in the P4 program. The Read RPC should work fine then. You can ignore the second sentence for now.

wmohsin: I guess folks here are OK with representation of counter with-respect-to default or uninitialized representation ... 0 for all fields on Read, which agrees with not setting the counter_data() field on Write (since 0 is default for proto3).

For meter, default behavior of tagging all packets GREEN makes sense. As for how to represent it, 0 for all fields doesn't make sense to me (since that says the rate is zero => RED). What do folks here think about using "-1" for all fields to indicate "no rate limiting" (i.e. all packets GREEN, which is the default behavior).
All fields in MeterConfig are int64, so -1 can be proposed as the "default" state.
What do you think?

antoninbas: Sounds good to me. So we have:
1. the default configuration for a meter is both rates and both burst sizes set to -1, which implies all packets are tagged GREEN.
2. when an entry is added to a table with a direct meter array and no configuration is provide, the default configuration will be used to program the hardware and it will be the configuration returned when doing a Read operation.
3. similarly, for an indirect meter array, all meters in the array will be initialized to the default configuration.
4. Any other meter configuration where one of the rates or one of the burst sizes is a negative number is illegal and will return an error.
5. Any meter configuration where the peak information rate is not greater than or equal to the committed information rate is an error.

wmohsin: Looks good.
(Adding similar statements in the p4 spec doc for counter ... where proto3 default of zero works out fine ... would be great)

antoninbas:
@aghaffarkhah just a remainder for the switch team to review this; please make sure you read the discussion comments

proto/p4/p4runtime.proto
```diff
@@ -136,6 +136,16 @@ message TableEntry {
...
+  // updating a table entry, leaving these fields unset means that the direct
+  // resources (if any) will not be updated. When reading a table entry, the
+  // values associated with the direct resources (if any) will be returned. If
+  // the table does not have a direct resource, then setting the direct resource
```
aghaffarkhah: Instead of "If the table does not have a direct resource" I would say "If the table does not support direct resources"

PR #190, opened 2017-08-19
In gNMI Subscribe return an error only on writes

antoninbas:


antoninbas:
@yyetim can you check if that works for you; I am no longer returning an error unless the client tries to subscribe to something

PR #201, opened 2017-09-07
Fix missing action profile group type

TakeshiTseng:
Action profile manager didn't store type of action profile group, so the
control plane gets UNSPECIFIED type (type 0) every time.


TakeshiTseng:
Thanks @antoninbas

PR #202, opened 2017-09-11
added clarification regarding default values for direct meter config …

samar-abdi:
…and counter data.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -139,12 +139,17 @@ message TableEntry {
   // meter_config and counter_data are convenience fields that enable the
   // controller to configure the direct resources associated with the table at
   // the same time as a match-action entry is inserted or modified. When
-  // updating a table entry, leaving these fields unset means that the direct
-  // resources (if any) will not be updated. When reading a table entry, the
-  // values associated with the direct resources (if any) will be returned. If
-  // the table does not support a direct resource, then setting the
-  // corresponding direct resource field in any table write operation will
-  // return an error.
+  // inserting a new table entry, leaving these fields unset means that the
+  // direct resources of this table (if any) will assume default values. For
+  // counters, the default value is 0, and for meters, the default value is
+  // always green. When updating a table entry, leaving these fields unset means
+  // that the direct resources (if any) will not be updated. When reading a table
+  // entry, the values associated with the direct resources (if any) will be
+  // returned. If these fields were never set or if the table does not support
+  // the direct resource, then these fields will be absent in the read table
+  // entry. If the table does not support a direct resource, then setting the
+  // corresponding direct resource field in any table write operation will return
+  // an error.
   MeterConfig meter_config = 6;
   CounterData counter_data = 7;
 }
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -252,7 +252,9 @@ message MeterEntry {
 }
 
 //------------------------------------------------------------------------------
-// For WriteRequest, Update.Type must be MODIFY.
+// For WriteRequest, Update.Type must be MODIFY. INSERT and DELETE on direct
+// meters is not allowed and will return an error. The insertion/deletion
+// should happen as part of INSERT/DELETE on the associated table-entry.
 // For ReadRequest, the scope is defined as follows:
 // - All meter cells for all meters if meter_id = 0 (default).
 // - All meter cells for given meter_id if table_entry.match is empty.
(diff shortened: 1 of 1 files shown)


proto/p4/p4runtime.proto
```diff
@@ -139,12 +139,17 @@ message TableEntry {
...
-  // corresponding direct resource field in any table write operation will
-  // return an error.
+  // inserting a new table entry, leaving these fields unset means that the
+  // direct resources of this table (if any) will assume default values. For
+  // counters, the default value is 0, and for meters, the default value is
+  // always green. When updating a table entry, leaving these fields unset means
+  // that the direct resources (if any) will not be updated. When reading a table
+  // entry, the values associated with the direct resources (if any) will be
+  // returned. If these fields were never set or if the table does not support
+  // the direct resource, then these fields will be absent in the read table
```
antoninbas: I think meter vs counter is different here. While this statement is true for meters, it's only true for counters if the table doesn't have a counter in the P4 program. Otherwise, if the table does have a counter, `counter_data` will always be set. This is because unlike the `meter_config`, `counter_data` is modified "by the data-plane".

proto/p4/p4runtime.proto
```diff
@@ -139,12 +139,17 @@ message TableEntry {
...
-  // values associated with the direct resources (if any) will be returned. If
-  // the table does not support a direct resource, then setting the
-  // corresponding direct resource field in any table write operation will
-  // return an error.
+  // inserting a new table entry, leaving these fields unset means that the
+  // direct resources of this table (if any) will assume default values. For
+  // counters, the default value is 0, and for meters, the default value is
+  // always green. When updating a table entry, leaving these fields unset means
+  // that the direct resources (if any) will not be updated. When reading a table
+  // entry, the values associated with the direct resources (if any) will be
```
wmohsin: I believe we discussed/agreed that the meter and counter case was different here:
- CounterData will always be set (could be 0s), unless table-definition doesn't have a direct counter
- MeterConfig will bet set only if table-definition has a direct meter, and it is not at its default configuration

proto/p4/p4runtime.proto
```diff
@@ -139,12 +139,17 @@ message TableEntry {
...
+  // returned. If these fields were never set or if the table does not support
+  // the direct resource, then these fields will be absent in the read table
+  // entry. If the table does not support a direct resource, then setting the
+  // corresponding direct resource field in any table write operation will return
+  // an error.
```
wmohsin: (Optional) May be all of this can be summarized into separate WRITE/READ sections like:

WRITE
- Return error if a direct resource field is set, but the table does not define the corresponding direct resource
- INSERT: "Not set" means the resource assume default value.
- MODIFY: "Not set" means the resource will not be updated

READ
- "Not set" if the table does not define the direct resource
- CounterData will always be set (could be 0s), unless table-definition doesn't have a direct counter
- MeterConfig will bet set only if table-definition has a direct meter, and it is not at its default configuration

proto/p4/p4runtime.proto
```diff
@@ -138,13 +138,24 @@ message TableEntry {
...
   // controller to configure the direct resources associated with the table at
-  // the same time as a match-action entry is inserted or modified. When
-  // updating a table entry, leaving these fields unset means that the direct
-  // resources (if any) will not be updated. When reading a table entry, the
-  // values associated with the direct resources (if any) will be returned. If
-  // the table does not support a direct resource, then setting the
-  // corresponding direct resource field in any table write operation will
-  // return an error.
+  // the same time as a match-action entry is inserted or modified.
+  // Table write:
```
wmohsin: This part is LGTM. I suggest also adding comments to clarify important related semantics, i.e. INSERT and DELETE on direct resources is not allowed and will return an error. The creation/deletion (or initializion/reset) should happen as part of INSERT/DELETE on the associated table-entry (I believe this was also part of the consensus).
For clarity, I suggest adding this comment on both DirectMeterEntry as well as DirectCounterEntry messages.

samar-abdi: Added the comments on the DirectMeterEntry as well as DirectCounterEntry messages as per suggestion. PTAL. Not sure if we should say anything about initialization/reset. The resource is inserted and initialized as part of TableEntry insertion, and we can modify it (including reset) using either TableEntry or  DirectMeterEntry/DirectCounterEntry.

proto/p4/p4runtime.proto
```diff
@@ -241,7 +252,9 @@ message MeterEntry {
 }
 
 //------------------------------------------------------------------------------
-// For WriteRequest, Update.Type must be MODIFY.
+// For WriteRequest, Update.Type must be MODIFY. INSERT and DELETE on direct
+// meters is not allowed and will return an error. The insertion/deletion
+// should happen as part of INSERT/DELETE on the associated table-entry.
```
antoninbas: nit: I would replace "should happen" with "happens"

proto/p4/p4runtime.proto
```diff
@@ -284,7 +297,9 @@ message CounterEntry {
 }
 
 //------------------------------------------------------------------------------
-// For WriteRequest, Update.Type must be MODIFY.
+// For WriteRequest, Update.Type must be MODIFY. INSERT and DELETE on direct
+// counters is not allowed and will return an error. The insertion/deletion
+// should happen as part of INSERT/DELETE on the associated table-entry.
```
antoninbas: same as above

PR #208, opened 2017-10-02
Make p4runtime.md draft public

yyetim:


proto/p4/p4runtime.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Value formats and constraints for P4Runtime fields
+
+This section describes some requirements on the values the fields in the
+P4Runtime can have. This is to ensure a canonical representation as described
+earlier. The server should handle fields as follows (using p4runtime.proto
+format):
+
+### `TableEntry.match`, `FieldMatch.field_id`, and other fields of the
+`FieldMatch`
```
antoninbas: I've just noticed this. You cannot split a header across 2 lines like this or the formatting is wrong with the Github markdown rendered. Would be great if you could amend your commit to remove the line break.

antoninbas:
I made a few small changes, and moved the spec under a docs folder; I hope there will be no objections. I'm going to merge it now since we wanted it on master by Tuesday, but we can revisit the location later.

PR #210, opened 2017-10-13
Added default action update flag in TableEntry.

samar-abdi:


samar-abdi:
Ugh. Sorry!

samar-abdi:
Incorporated Antonin's and Tom's feedback. PTAL.

proto/docs/p4runtime.md
```diff
@@ -126,7 +142,22 @@ table_entry {
 
 If `match` for a field is not defined, then the field is treated as don't care
 when constructing the match key for the table entry. Note that leaving match field
-undefined is only allowed for ternary and LPM fields.
+undefined is only allowed for ternary and LPM fields. The following table entry
```
teverman: The old code allowed for the possibility of don't care range matches - see lines 159-160 of the old code.  Are we back to not allowing don't care range matches?

proto/docs/p4runtime.md
```diff
@@ -108,15 +108,31 @@ format):
...
-of a table:
+to true when it wants to change the default action within a table. The following
+table illustrates this use case. The ids in the comments refer to the P4 compiler
+assigned ids to the entities. We assume that both actions `bar` and `baz` do not
+take any arguments.
 
 ```
+table foo {             // table id: 1
+  key = {
+    hdr.foo : ternary;  // field id: 10
```
teverman: Missing closing } for key.

proto/docs/p4runtime.md
antoninbas: I think that's just an omission

antoninbas:
@samar-abdi @teverman I'll merge this as soon as the regression passes

PR #212, opened 2017-10-19
Fix pi_act_prof_grps_next method

TakeshiTseng:
The `curr_group` does not point to next group after `pi_act_prof_grps_next`
called.


antoninbas:
@TakeshiTseng Thanks for tracking this down BTW

PR #217, opened 2017-10-24
Support is_default_action flag in P4 Runtime implementation

antoninbas:
Default action is no longer indicated by an empty match key.


antoninbas:
@teverman @samar-abdi FYI I have updated my P4 Runtime implementation to use the new `is_default_action` flag in `TableEntry`

PR #219, opened 2017-10-30
some cpp fe accessors

melshabani:


frontends_extra/cpp/src/tables.cpp
```diff
@@ -204,6 +209,11 @@ MatchKey::from(const pi_match_key_t *pi_match_key) {
   memcpy(match_key->data, pi_match_key->data, mk_size);
 }
 
+pi_p4_id_t
+MatchKey::get_table_id() const {
+    return reader.get_table_id();
```
antoninbas: I wouldn't use the `reader` for this

frontends_extra/cpp/src/tables.cpp
```diff
@@ -501,6 +511,10 @@ ActionData::reset() {
   nset = 0;
 }
 
+pi_p4_id_t ActionData::get_action_id() const{
+    return reader.get_action_id();
```
antoninbas: same

antoninbas:
@melshabani style check is failing

PR #222, opened 2017-11-01
Update p4runtime.proto

wmohsin:
Align {Set|Get}ForwardingPipelineConfig RPCs with the rest of the API, which is on a per device basis. This also simplifies error reporting by simply using the top-level grpc status returned. The changes are backwards-compatible, and older fields will be removed once all clients have migrated away from them.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -393,25 +393,39 @@ message SetForwardingPipelineConfigRequest {
     VERIFY_AND_COMMIT = 3;
     COMMIT = 4;
   }
-  Uint128 election_id = 1;
-  repeated ForwardingPipelineConfig configs = 2;
+  uint64 device_id = 1;
+  Uint128 election_id = 2;
   Action action = 3;
+  ForwardingPipelineConfig config = 4;
+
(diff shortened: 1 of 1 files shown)


proto/p4/p4runtime.proto
```diff
@@ -393,25 +393,39 @@ message SetForwardingPipelineConfigRequest {
...
-  Uint128 election_id = 1;
-  repeated ForwardingPipelineConfig configs = 2;
+  uint64 device_id = 1;
+  Uint128 election_id = 2;
   Action action = 3;
+  ForwardingPipelineConfig config = 4;
+
+  // TODO(wmohsin): Remove when clients have migrated to 'device_id' and
+  // 'config' above.
+  repeated ForwardingPipelineConfig configs = 2 [deprecated = true];
```
antoninbas: you cannot use the same protobuf field id twice (2), even for a deprecated field, it is causing the build to fail
apart from this, looks good to me

aghaffarkhah: Yes the field id needs to change.

wmohsin: Done

PR #223, opened 2017-11-01
Update p4runtime.proto

wmohsin:
Removes ActionProfileGroup.type field. In the early P4Runtime days, we added an APG 'type' in a somewhat half-baked way, with the understanding that we will revisit/refine it down the road. We have now come to a consensus that this is a property of the P4 action-selector object, and should not be captured here.


antoninbas:
@wmohsin The tests are failing because of some code using this attribute. I'll update the code.

antoninbas:
FYI, waiting on @aghaffarkhah or @teverman review to merge this

PR #224, opened 2017-11-02
Use new {Set|Get}ForwardingPipelineConfig RPC fields

antoninbas:
p4runtime.proto was updated and the format of the
{Set|Get}ForwardingPipelineConfig RPCs was modified, with some fields
marked as deprecated and some new fields introduced. This commit makes
sure that only the new fields are used.


antoninbas:
See #222

antoninbas:
@teverman @samar-abdi 
FYI I'm merging this, which means if you start using a version of PI / bmv2 which includes this commit, you will need to be using the new fields, not the deprecated ones.

PR #226, opened 2017-11-03
Add example PTF test for P4 Runtime

antoninbas:
We add some infrastructure code to easily run PTF tests written with P4
Runtime accross different targets. The objective is to have a testsuite
which can be used to evaluate conformance (and potentially to benchmark)
different P4 Runtime implementations.

As an example, we provide a first P4 program, l3_fwd_host.p4, along with
a test. We also provide instructions on how to run this PTF test on the
bmv2 simple_switch_grpc target.


proto/ptf/l3_host_fwd/test/l3_host_fwd.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # no forwarding entry: packet must be dropped
+        testutils.send_packet(self, ig_port, pkt)
+        testutils.verify_no_other_packets(self)
+
+        # add a forwarding entry
+        self.send_request_add_entry_to_action(
+            "l3_host_fwd", [self.Exact("hdr.ipv4.dst_addr", ip_dst_addr_str)],
+            "set_nexthop",
+            [("port", eg_port_str), ("smac", smac), ("dmac", dmac)])
```
jafingerhut: Does this method call return any kind of error status that could be used to do negative tests for API calls that should fail?

antoninbas: This method will throw an exception if the server fails to process the request (error codes will be standardized as part of the spec). The best thing to do if you expect an error may be to catch it yourself and assert the test if the exception is not raised. I'll try to add an example later.

antoninbas: @jafingerhut I have added the necessary infrastructure to perform negative tests in a convenient way. I have also added a couple examples illustrating how to use it.

Issue #228, opened 2017-11-13
Make well, but Make Check failed.


MaoJianwei:
Steps:

git clone https://github.com/p4lang/PI.git
git submodule update --init --recursive
./autogen.sh
./configure --with-bmv2 --with-proto
sudo make
sudo make check

and....
This is the log, thank you!

```

======================================
   PI 0.1: tests/CLI/test-suite.log
======================================

# TOTAL: 13
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  13
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: table_dump
================

Traceback (most recent call last):
  File "./run_one_test.py", line 219, in <module>
    main()
  File "./run_one_test.py", line 134, in main
    stderr=subprocess.PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
FAIL table_dump.test (exit status: 1)

FAIL: table_dump_valid
======================

Traceback (most recent call last):
  File "./run_one_test.py", line 219, in <module>
    main()
  File "./run_one_test.py", line 134, in main
    stderr=subprocess.PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
FAIL table_dump_valid.test (exit status: 1)

FAIL: table_indirect
====================

Traceback (most recent call last):
  File "./run_one_test.py", line 219, in <module>
    main()
  File "./run_one_test.py", line 134, in main
    stderr=subprocess.PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
FAIL table_indirect.test (exit status: 1)

FAIL: table_wkey
================

Traceback (most recent call last):
  File "./run_one_test.py", line 219, in <module>
    main()
  File "./run_one_test.py", line 134, in main
    stderr=subprocess.PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
[...]

PR #229, opened 2017-11-14
fix Dependencies - to make check successfully - for issue #228

MaoJianwei:
https://github.com/p4lang/PI/issues/228

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ To include `p4runtime.proto` in the build, please run `configure` with
 ./autogen.sh
 ./configure [--with-proto] [--with-bmv2]
 make
-sudo make check
+make check
 sudo make install
 ```
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ See [examples](examples/) for how to use the PI.
 
 - libjudy-dev
 - libreadline-dev
+
+## CI Tests Dependencies
+
 - libtool-bin
 - valgrind
 
--- a/README.md
+++ b/README.md
@@ -11,11 +11,21 @@ See [examples](examples/) for how to use the PI.
 
 - libjudy-dev
 - libreadline-dev
+- libtool-bin
+- valgrind
 
 ## Building p4runtime.proto
 
 To include `p4runtime.proto` in the build, please run `configure` with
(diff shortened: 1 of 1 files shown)


MaoJianwei:
all of existed CI tests can pass successfully.

README.md
```diff
@@ -11,12 +11,22 @@ See [examples](examples/) for how to use the PI.
 
 - libjudy-dev
 - libreadline-dev
+- libtool-bin
```
antoninbas: These should probably be listed as `make check` dependencies. They are not needed except to run the full test suite when configure is run with `--with-bmv2`

README.md
```diff
@@ -11,12 +11,22 @@ See [examples](examples/) for how to use the PI.
...
 ## Building p4runtime.proto
 
 To include `p4runtime.proto` in the build, please run `configure` with
 `--with-proto`.
 
+```
+./autogen.sh
+./configure [--with-proto] [--with-bmv2]
+make
+sudo make check
```
antoninbas: `sudo` should not be needed here

MaoJianwei: nope, SUDO make check is necessary. or nanomsg ipc tests will fail

README.md
MaoJianwei: yes, and I fix it, could you please review again? thanks!

MaoJianwei:
SUDO make check is necessary, or nanomsg ipc tests will fail

README.md
antoninbas: This is not correct, unless you have some left-over ipc files which were created with root privileges. That is because nanomsg doesn't unlink the files when `nn_close` is called (see https://github.com/nanomsg/nanomsg/issues/666).
I suggest you try running `sudo rm -rf /tmp/*.ipc`. After that you will be able to run `make check` without `sudo` any number of times.

MaoJianwei: Yes, you are right. I remove *.ipc and it works.
I have corrected the line, could you please review again, thank you!

MaoJianwei:
Yes, you are right. I remove *.ipc and it works.
I have corrected the line, could you please review again, thank you!

PR #238, opened 2017-11-21
Update gnmi.proto to latest version

antoninbas:
The proto file was moved to a new git repo (from openconfig/reference to
openconfig/gnmi).


antoninbas:
Users may need to run `git submodule sync` after pulling this change

PR #240, opened 2017-11-28
Added PIGrpcServerGetPort API to get back the port number being used by the server

dushyantarora13:
This is useful in testing when we start the server on a random port.

### DIFF
--- a/proto/tests/server/test_no_pipeline_config.cpp
+++ b/proto/tests/server/test_no_pipeline_config.cpp
@@ -80,7 +80,6 @@ class TestNoForwardingPipeline : public ::testing::Test {
 constexpr char TestNoForwardingPipeline::grpc_server_addr[];
 
 TEST_F(TestNoForwardingPipeline, Write) {
-  EXPECT_EQ(PIGrpcServerGetPort(), 50051);
   p4::WriteRequest request;
   request.set_device_id(device_id);
   ClientContext context;
--- a/proto/tests/server/test_no_pipeline_config.cpp
+++ b/proto/tests/server/test_no_pipeline_config.cpp
@@ -80,6 +80,7 @@ class TestNoForwardingPipeline : public ::testing::Test {
 constexpr char TestNoForwardingPipeline::grpc_server_addr[];
 
 TEST_F(TestNoForwardingPipeline, Write) {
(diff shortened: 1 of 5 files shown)


proto/tests/server/test_no_pipeline_config.cpp
```diff
@@ -80,6 +80,7 @@ class TestNoForwardingPipeline : public ::testing::Test {
 constexpr char TestNoForwardingPipeline::grpc_server_addr[];
 
 TEST_F(TestNoForwardingPipeline, Write) {
+  EXPECT_EQ(PIGrpcServerGetPort(), 50051);
```
antoninbas: Thanks for adding a test but please write a separate test case for this.
Adding this in an existing test case in this file doesn't make sense IMO. Maybe create a new file for that kind of generic server test (test_server.cpp?).

antoninbas:
In order for us to accept your PR, please confirm that your organization is part of p4.org.

dushyantarora13:
I work for Google (dushyantarora@google.com)

proto/tests/server/test_no_pipeline_config.cpp
dushyantarora13: Moved test to new file test_pi_server.cpp

proto/tests/server/test_pi_server.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <PI/proto/pi_server.h>
+#include <string>
+
+namespace pi {
+namespace proto {
+namespace testing {
+namespace {
+
+static std::string grpc_server_addr = "0.0.0.0";
+static std::string grpc_server_port = "50051";
```
antoninbas: please use `50054`. It's a little bit hacky at the moment but each test file has to use a different port number in case the tests are run in parallel. `50051...50053` are already taken by the other files.
long term, we definitely need a better way of allocating ports for testing. Either something like this (https://github.com/grpc/grpc/blob/master/test/core/util/port.h) or something simpler by creating unique temporary directories.

dushyantarora13: Fixed. Thanks for the pointer to port.h

PR #251, opened 2017-12-08
Multicast

samar-abdi:


proto/p4/p4runtime.proto
```diff
@@ -314,6 +315,33 @@ message CounterData {
...
+  oneof pre_entry {
+    MulticastGroupEntry multicast_group_entry = 1;
+    CloneSessionEntry clone_session_entry = 2;
+  }
+}
+
+// The egress instance must be unique for each replica in a given multicast
+// group entry.
+message MulticastGroupEntry {
+  bytes multicast_group = 1;
```
antoninbas: I thought we agreed this would be a `int64`. It makes it much easier to manipulate it in software. We can ask the PSA group to constraint the bitwidth to <= 64-bit for the corresponding metadata field.

proto/p4/p4runtime.proto
```diff
@@ -314,6 +315,33 @@ message CounterData {
...
+  message Replica {
+    bytes port = 1;
+    bytes egress_instance = 2;
```
antoninbas: `int64`, same as for `multicast_group`

proto/p4/p4runtime.proto
```diff
@@ -314,6 +315,33 @@ message CounterData {
...
+  }
+  repeated Replica replicas = 2;
+}
+
+message CloneSessionEntry {
+  bytes session_id = 1;
```
antoninbas: `int64`

proto/p4/p4runtime.proto
```diff
@@ -314,6 +315,33 @@ message CounterData {
...
+  bytes port = 2;
+  bytes class_of_service = 3;
```
antoninbas: `int64` as well IMO

proto/p4/p4runtime.proto
```diff
@@ -314,6 +315,33 @@ message CounterData {
...
+    CloneSessionEntry clone_session_entry = 2;
+  }
+}
+
+// The egress instance must be unique for each replica in a given multicast
+// group entry.
+message MulticastGroupEntry {
+  bytes multicast_group = 1;
+  message Replica {
+    bytes port = 1;
```
antoninbas: As discussed in person, I believe the notion of port may be too restrictive for this (as for the action profile member watch attribute). I would keep the `bytes` type but rename it (maybe `destination`?). We should discuss this topic as the next P4 API WG meeting, along with the watch attribute,

proto/p4/p4runtime.proto
samar-abdi: Ya, I thought about it and put it on hold. We are not doing this for match fields or action parameters, so a multicast group as action parameter would be bytes, but as a PRE entry field would be int64. It seems a little inconsistent, but I agree it is a pain to manipulate bytes in SW and having to take care of endianness etc.

antoninbas: This is not an arbitrary user-defined field though, this is an architecture-defined field. In a way we are hard-coding a name for it, so we can also hard-code a type for it.

proto/p4/p4runtime.proto
samar-abdi: Yes, let's discuss this further. IMHO, it would be cleaner to draw a boundary and keep everything in the P4Runtime service at the switch scope: hence the notion of singletons and trunks is OK, but no higher-level abstractions like routes. If we want to keep a watch on routes, that would need to be managed by the controller, not the P4Runtime service.

antoninbas: Neither P4 nor PSA give special status to LAGs. So it should be either singleton hardware ports only (which do have special status) or a generic mechanism that is able to support LAG among other protocols / abstractions. I'm for consistency.

samar-abdi: PSA defines the architecture of a reconfigurable switching ASIC. If we were to treat P4Runtime as a runtime API to a PSA-targeted P4 program, then it will turn into a very thin SDK of a reconfigurable switching ASIC. At which point, it will be not be very useful to an SDN controller, that needs an interface to a switch, not ASIC. I am hoping we can settle on the right abstraction and scope of P4Runtime like the PSA group did. Let's discuss this more at the WG meeting. This dichotomy between switch and ASIC is the source of endless confusion.

proto/p4/p4runtime.proto
samar-abdi: Fair point. In that case, we should make all fields uint64 (including ports).

proto/p4/p4runtime.proto
samar-abdi: I am going to flip-flop here and propose that the egress port should be a singleton by default for both multicast and clone cases. If we want to do translation/resolution/watch in the p4runtime service, we should do so through annotations.

antoninbas:
@wmohsin please update your review if you're happy with the changes
@teverman you may want to take a look as well

teverman:
LGTM

PR #254, opened 2017-12-12
Do not assert on read when priority set for exact

antoninbas:
The priority value in the TableEntry message only has meaning for
ternary matches. Targets may not store the unsolicited priority value
when one is provided for a non-ternary table - this is the case of
bmv2 -, which means the priority value may not be returned by a table
read operation. The P4 Runtime server implementation expects the target
to return entries which are identical to the ones written, which was
causing a crash. For now, we simply ignore unsolicited priority
values. A better fix may have been to ignore the priority when comparing
match keys for non-ternary tables in the table_info_store, but the
required changes would have been more complex.

The best thing to do would probably be to return an error for
unsolicited priority values, which requires updating the P4 Runtime
specification.

We added a gtest to make sure the crash was fixed: it required making
the table implementation in the mock switch aware of the table match
type, which is why this commit is quite large.

### DIFF
--- a/proto/frontend/src/device_mgr.cpp
+++ b/proto/frontend/src/device_mgr.cpp
@@ -1122,6 +1122,7 @@ class DeviceMgrImp {
     auto status = validate_match_key(entry);
     if (IS_ERROR(status)) return status;
     auto t_id = entry.table_id();
+    bool need_priority = false;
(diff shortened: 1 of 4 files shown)


antoninbas:
See https://github.com/p4lang/behavioral-model/issues/490

PR #262, opened 2017-12-18
Return error in P4 Runtime if priority > 0 for non-TCAM

### LINKED ISSUE #253: Should providing a non-zero (i.e. non-default) priority value for a non-ternary match-table be an error?
The priority field should only be used when the match table includes a ternary / range match.
Should we silently ignore the priority are fail with an INVALID_ARGUMENT error?

antoninbas:
fixes #253


antoninbas:
@wmohsin @samar-abdi please review addition to P4Runtime spec

PR #263, opened 2017-12-21
Add openconfig-if-ethernet.yang to supported models

antoninbas:
This YANG model augments openconfig-interfaces for ethernet interfaces.
This commit includes a work-around to bypass a sysrepo
restriction. sysrepo does not seem to be supporting "when" statements
involving operational state in the YANG. Therefore when loading the YANG
model into sysrepo, we do a sed to replace `when
"oc-if:state/oc-if:type"` with `when "oc-if:config/oc-if:type"`/


antoninbas:
See https://github.com/sysrepo/sysrepo/issues/1018 for related sysrepo issue

PR #270, opened 2017-12-29
Remove the old demo code using internal RPC

antoninbas:
This code was meant to demonstrate the internal RPC mechanism which used
Nanomsg with a custom message serialization format. We are now promoting
p4runtime.proto as the preferred RPC mechanism for P4 Runtime (there is
an equivalent demo under proto/demo_grpc). We are removing the old demo
code to avoid having to maintain it and to avoid misleading newcomers to
the P4 Runtime project.


antoninbas:
Fixes #25

PR #272, opened 2018-01-07
Added new APIs PIGrpcServerGetPacketInCount and PIGrpcServerGetPacketOutCount for testing and debugging

dushyantarora:
The APIs return the number of StreamMessageResponse packets sent to grpc client and PacketOut messages sent to DeviceMgr respectively

### DIFF
--- a/proto/server/PI/proto/pi_server.h
+++ b/proto/server/PI/proto/pi_server.h
@@ -33,6 +33,12 @@ void PIGrpcServerRunAddr(const char *server_address);
 // Get port number bound to the server
 int PIGrpcServerGetPort();
 
+// Get number of PacketIn packets sent to client
+uint64_t PIGrpcServerGetPacketInCount(uint64_t device_id);
+
+// Get number of PacketOut packets sent to DevMgr
+uint64_t PIGrpcServerGetPacketOutCount(uint64_t device_id);
+
 // Wait for the server to shutdown. Note that some other thread must be
 // responsible for shutting down the server for this call to ever return.
 void PIGrpcServerWait();
--- a/proto/server/pi_server.cpp
+++ b/proto/server/pi_server.cpp
@@ -172,7 +172,7 @@ class DeviceState {
     return device_mgr.get();
   }
 
-  void send_packet_in(p4::PacketIn *packet) const {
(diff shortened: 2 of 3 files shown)


proto/server/PI/proto/pi_server.h
```diff
@@ -17,6 +17,9 @@
  * Antonin Bas (antonin@barefootnetworks.com)
  *
  */
+#include <PI/frontends/proto/device_mgr.h>
```
antoninbas: `PI/frontends/proto/device_mgr.h` is a C++ header, you cannot include it in this header which is meant to be usable by both C++ and C compilers (which is why as you can see, there is no `PIGrpcServer` class).
Since you are only including it for the definition of `device_id_t` and nothing else, I would not bother and simply hardcode it to 64-bit.

proto/server/PI/proto/pi_server.h
```diff
@@ -33,6 +36,9 @@ void PIGrpcServerRunAddr(const char *server_address);
 // Get port number bound to the server
 int PIGrpcServerGetPort();
 
+// Get number of PacketIn packets sent to client
+uint64_t PIGrpcServerGetPacketInCount(DeviceMgr::device_id_t device_id);
```
antoninbas: Why not kill 2 birds with 1 stone and support packet-out counters with the same function?
Also, the current behavior is to "create" a device if the device_id does not exist yet and return a count of zero. Is this the desired behavior or should we return some kind of error?

proto/server/PI/proto/pi_server.h
dushyantarora13: Fixed

proto/server/PI/proto/pi_server.h
dushyantarora13: I added new API PIGrpcServerGetPacketOutCount. I thought that would be neater. Fixed current behavior to not create DeviceState if a device_id does not exist. I think it is fine to return 0 in case device_id does not exist yet.  Technically, we didn't receive/send any packets from/to non-existent device.

dushyantarora:
I plan to add more tests for the APIs in simple_switch_grpc target once this PR merges.

PR #279, opened 2018-01-16
Stricter match field format for ternary matches

antoninbas:
We require that no unused bit be set in ternary (and LPM) match field
values. If this is violated, the P4 Runtime server must return an
INVALID_ARGUMENT error.

### DIFF
--- a/proto/docs/p4runtime.md
+++ b/proto/docs/p4runtime.md
@@ -178,9 +178,13 @@ switch(p4_type(match.field_id())) {
   case TERNARY:
     assert(match.has_ternary())
     assert(!match.ternary().value().empty() && !match.ternary().mask().empty())
+    assert(match.ternary().value() ==
+           (match.ternary().value() & match.ternary().mask()))
   case LPM:
     assert(match.has_lpm())
     assert(!match.lpm().value().empty() && match.lpm().prefix_len() > 0)
+    assert(count_trailing_zeros(match.lpm().value()) >=
+           p4_bitwidth(match.field_id()) - match.lpm().prefix_len())
   case RANGE:
     Integer low, high;
     assert(match.has_range())
--- a/proto/docs/p4runtime.md
+++ b/proto/docs/p4runtime.md
@@ -176,14 +176,16 @@ switch(p4_type(match.field_id())) {
   case EXACT:
     assert(match.has_exact() && !match.exact().value().empty())
   case TERNARY:
+    Integer value, mask;
     assert(match.has_ternary())
     assert(!match.ternary().value().empty() && !match.ternary().mask().empty())
-    assert(match.ternary().value() ==
-           (match.ternary().value() & match.ternary().mask()))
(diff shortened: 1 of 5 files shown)


proto/docs/p4runtime.md
```diff
@@ -178,9 +178,13 @@ switch(p4_type(match.field_id())) {
   case TERNARY:
     assert(match.has_ternary())
     assert(!match.ternary().value().empty() && !match.ternary().mask().empty())
+    assert(match.ternary().value() ==
+           (match.ternary().value() & match.ternary().mask()))
```
teverman: In accordance with the conventions above, shouldn't this be:
 assert(parseInteger(match.ternary().value()) ==
￼            (parseInteger(match.ternary().value()) & parseInteger(match.ternary().mask())))

proto/frontend/src/common.cpp
```diff
@@ -40,6 +40,14 @@ uint8_t clz(uint8_t b) {
   return (hb0 == 0) ? (4 + clz_table_hb[hb1]) : clz_table_hb[hb0];
 }
 
+uint8_t ctz(uint8_t b) {
+  static constexpr uint8_t ctz_table_hb[16] =
+      {4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0};
+  uint8_t hb0 = b >> 4;
+  uint8_t hb1 = b & 0x0f;
```
teverman: Nit: I would prefer more readable names like high_nibble and low_nibble, but this conforms to existing convention within this file.

proto/frontend/src/common.cpp
```diff
@@ -51,6 +59,17 @@ Code check_proto_bytestring(const std::string &str, size_t nbits) {
   return Code::OK;
 }
 
+bool check_prefix_trailing_zeros(const std::string &str, int pLen) {
+  size_t trailing_zeros = str.size() * 8 - pLen;
```
teverman: This looks like it would behave badly given input where str.size() * 8 < pLen

antoninbas:
Thanks for the review @teverman! I tries to capture your feedback in a new commit.

proto/docs/p4runtime.md
```diff
@@ -176,11 +176,17 @@ switch(p4_type(match.field_id())) {
   case EXACT:
     assert(match.has_exact() && !match.exact().value().empty())
   case TERNARY:
+    Integer value, mask;
     assert(match.has_ternary())
     assert(!match.ternary().value().empty() && !match.ternary().mask().empty())
+    assert(parseInteger(match.ternary().value(), &value))
+    assert(parseInteger(match.ternary().mask(), &mask))
+    assert(value == value & mask)
```
teverman: Doesn't == have higher precedence than bitwise & ?

antoninbas: Thanks for spotting this; I mistakenly removed the parenthesis when I made the `parseInteger` change.

PR #280, opened 2018-01-16
Removed counter_id and meter_id fields from DirectCounterEntry and

### LINKED ISSUE #266: Remove counter id and meter id for direct counter/meter entry message
Direct counters are addressed by a table entry. But there is a counter_id in those messages to read all cells of a direct counter. However, since the implication is that there is only one direct counter in a table, couldn't setting the table id in the table_entry (field 2) achieve the same result?

Same for direct meters.

The concern here is that the existing DirectCounterEntry message conflicts with the way of reading direct counters as part of the table entry read. There is no counter_id in the latter.

dushyantarora:
DirectMeterEntry respectively. Fixes #266

### DIFF
--- a/proto/frontend/src/device_mgr.cpp
+++ b/proto/frontend/src/device_mgr.cpp
@@ -475,10 +475,13 @@ class DeviceMgrImp {
   Status direct_meter_write(p4::Update_Type update,
                             const p4::DirectMeterEntry &meter_entry,
                             const SessionTemp &session) {
-    if (!check_p4_id(meter_entry.meter_id(), P4ResourceType::METER))
-      return make_invalid_p4_id_status();
-
+    if (!meter_entry.has_table_entry()) {
+      RETURN_ERROR_STATUS(Code::INVALID_ARGUMENT,
+                          "Missing table_entry field in DirectMeterEntry");
+    }
     const auto &table_entry = meter_entry.table_entry();
+    if (!check_p4_id(table_entry.table_id(), P4ResourceType::TABLE))
+      return make_invalid_p4_id_status();
     auto table_lock = table_info_store.lock_table(table_entry.table_id());
 
     pi_entry_handle_t entry_handle = 0;
@@ -487,6 +490,12 @@ class DeviceMgrImp {
       if (IS_ERROR(status)) return status;
     }
 
+    p4_id_t table_direct_meter_id = pi_get_table_direct_resource_p4_id(
+        table_entry.table_id(), P4ResourceType::METER);
+    if (table_direct_meter_id == PI_INVALID_ID) {
+      RETURN_ERROR_STATUS(Code::INVALID_ARGUMENT,
+                          "Table has no direct meters");
+    }
     switch (update) {
       case p4::Update_Type_UNSPECIFIED:
         RETURN_ERROR_STATUS(Code::INVALID_ARGUMENT);
@@ -497,7 +506,7 @@ class DeviceMgrImp {
         {
           auto pi_meter_spec = meter_spec_proto_to_pi(meter_entry.config());
           auto pi_status = pi_meter_set_direct(session.get(), device_tgt,
-                                               meter_entry.meter_id(),
+                                               table_direct_meter_id,
                                                entry_handle,
                                                &pi_meter_spec);
           if (pi_status != PI_STATUS_SUCCESS) {
@@ -511,7 +520,7 @@ class DeviceMgrImp {
           pi_meter_spec_t pi_meter_spec =
               {0, 0, 0, 0, PI_METER_UNIT_DEFAULT, PI_METER_TYPE_DEFAULT};
           auto pi_status = pi_meter_set_direct(session.get(), device_tgt,
-                                               meter_entry.meter_id(),
+                                               table_direct_meter_id,
                                                entry_handle,
                                                &pi_meter_spec);
           if (pi_status != PI_STATUS_SUCCESS) {
@@ -961,15 +970,10 @@ class DeviceMgrImp {
     RETURN_OK_STATUS();
   }
 
-  Status direct_counter_read_one(p4_id_t counter_id,
-                                 const p4::DirectCounterEntry &counter_entry,
+  Status direct_counter_read_one(const p4::TableEntry &table_entry,
                                  const SessionTemp &session,
                                  p4::ReadResponse *response) const {
-    assert(pi_p4info_counter_get_direct(p4info.get(), counter_id) !=
-           PI_INVALID_ID);
-
-    if (counter_entry.has_table_entry()) {
-      const auto &table_entry = counter_entry.table_entry();
+    if (table_entry.match_size() > 0) {
       auto table_lock = table_info_store.lock_table(table_entry.table_id());
 
       pi_entry_handle_t entry_handle = 0;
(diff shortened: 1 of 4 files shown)


proto/frontend/src/device_mgr.cpp
```diff
@@ -475,10 +475,17 @@ class DeviceMgrImp {
...
                             const SessionTemp &session) {
-    if (!check_p4_id(meter_entry.meter_id(), P4ResourceType::METER))
-      return make_invalid_p4_id_status();
-
+    if (!meter_entry.has_table_entry()) {
+      RETURN_ERROR_STATUS(Code::UNIMPLEMENTED,
+                          "Modifying ALL direct meters is not supported yet");
+    }
     const auto &table_entry = meter_entry.table_entry();
+    if (table_entry.table_id() == 0) {
```
antoninbas: I don't think we should have 2 ways of "modifying all direct meters": missing `table_entry` (default) and `table_entry` with a default `table_id` of 0. Maybe we should agree that `table_entry` is always required?

proto/frontend/src/device_mgr.cpp
```diff
@@ -475,10 +475,17 @@ class DeviceMgrImp {
...
+      RETURN_ERROR_STATUS(Code::UNIMPLEMENTED,
+                          "Modifying ALL direct meters is not supported yet");
+    }
```
antoninbas: what about the case where `table_entry.match_size() == 0`

proto/frontend/src/device_mgr.cpp
```diff
@@ -999,27 +1014,18 @@ class DeviceMgrImp {
...
-          "When reading direct counters, you cannot use a counter_id of zero "
-          "with a non-empty table_entry");
+    if (!counter_entry.has_table_entry()) {
+      RETURN_ERROR_STATUS(Code::UNIMPLEMENTED,
+                          "Reading ALL direct counters is not supported yet");
     }
-    if (counter_id == 0) {
+    const auto &table_entry = counter_entry.table_entry();
+    if (table_entry.table_id() == 0) {
       RETURN_ERROR_STATUS(Code::UNIMPLEMENTED,
```
antoninbas: same as above, should we have 2 ways of doing this?

proto/p4/p4runtime.proto
```diff
@@ -255,10 +255,10 @@ message MeterEntry {
 // meters is not allowed and will return an error. The insertion/deletion
 // should happen as part of INSERT/DELETE on the associated table-entry.
 // For ReadRequest, the scope is defined as follows:
-// - All meter cells for all meters if meter_id = 0 (default).
-// - All meter cells for given meter_id if table_entry.match is empty.
+// - All meter cells for all tables if table_entry is empty (default).
+// - All meter cells of a table if table_entry.table_id is present and
+//   table_entry.match is empty.
```
antoninbas: you do not describe the case where `table_entry` is present and `table_entry.table_id == 0`. In the code you seem to treat this case as equivalent to an empty / missing `table_entry` (default), but as I wrote above, I wonder if we should have 2 ways of doing the same thing or if we should mandate that `table_entry` is required.

proto/p4/p4runtime.proto
```diff
@@ -255,10 +255,10 @@ message MeterEntry {
 // meters is not allowed and will return an error. The insertion/deletion
 // should happen as part of INSERT/DELETE on the associated table-entry.
 // For ReadRequest, the scope is defined as follows:
```
antoninbas: The comment only mentions `ReadRequest` but in the code you seem to imply that wildcards will also be supported in the long-term for `WriteRequest`. I am not opposed to this, but I think the comment here should mention `WriteRequest` as well if we intend to support this eventually.

proto/frontend/src/device_mgr.cpp
dushyantarora: Fine with me. Fixed.

proto/frontend/src/device_mgr.cpp
dushyantarora: Not allowing wildcard matching for now. entry_handle_from_table_entry(table_entry, &entry_handle); will return an INVALID_ARGUMENT error.

proto/frontend/src/device_mgr.cpp
dushyantarora: Fixed

proto/p4/p4runtime.proto
dushyantarora: Removed wildcard for WriteRequest. I don't know any use case for it. Treating missing table_entry as INVALID_ARGUMENT

proto/p4/p4runtime.proto
dushyantarora: No longer needed. table_id 0 will return an INVALID_ARGUMENT error.

proto/p4/p4runtime.proto
```diff
@@ -255,10 +255,10 @@ message MeterEntry {
...
-// - All meter cells for all meters if meter_id = 0 (default).
-// - All meter cells for given meter_id if table_entry.match is empty.
+// - All meter cells for all tables if table_entry is empty (default).
```
wmohsin: For style parity with other APIs here ("read all table entries", "read all indirect meters"), I suggest we require that table_entry is always present (i.e. has_table_entry() == true), and table_id=0 implying "all meter cells for all tables".

proto/tests/test_proto_fe.cpp
```diff
@@ -1530,7 +1505,19 @@ TEST_F(DirectCounterTest, InvalidRequestReadAllFromDefault) {
 
 // TODO(antonin)
 TEST_F(DirectCounterTest, ReadAllFromTable) {
-  auto counter_entry = make_counter_entry(nullptr);  // default TableEntry
+  std::string mf("\xaa\xbb\xcc\xdd", 4);
```
antoninbas: why do we need to add an actual table entry to test this?

proto/p4/p4runtime.proto
antoninbas: so to summarize:
- for a WriteRequest: table_entry is required, table_entry.table_id must be valid and the match key must be provided
- for a ReadRequest: default table_entry means "read-all entries in all direct counters / meters". Otherwise table_entry.table_id must be valid; in which case a missing match key means "read-all entries in the direct counter / meter for this specific table" and a valid match key means "read just this specific entry in the direct counter / meter for this specific table".

Please confirm.

dushyantarora:
There is only 1 modification here. I talked with Samar and Waqar and we converged on the following:

- for a WriteRequest: table_entry is required, table_entry.table_id must be valid and the match key must be provided (the same)
- for a ReadRequest: table_entry with table_id= 0 means "read-all entries in all direct counters / meters for all tables". Otherwise table_entry.table_id must be valid; in which case a missing match key means "read-all entries in the direct counter / meter for this specific table" and a valid match key means "read just this specific entry in the direct counter / meter for this specific table". Missing table_entry field will be treated as INVALID_ARGUMENT.

Will send out PR with changes soon.

proto/tests/test_proto_fe.cpp
dushyantarora: We don't. I thought I will keep it like this so it will easier to test when we actually implement it.

proto/p4/p4runtime.proto
```diff
@@ -255,10 +255,10 @@ message MeterEntry {
...
-// - All meter cells for all meters if meter_id = 0 (default).
-// - All meter cells for given meter_id if table_entry.match is empty.
+// - All meter cells for all tables if table_entry.table_id = 0.
+// - All meter cells of a table if table_entry.table_id is present and
+//   table_entry.match is empty.
 message DirectMeterEntry {
-  uint32 meter_id = 1;
   // The associated table entry match.
   // table_entry.action is ignored. Other fields specify the match.
   TableEntry table_entry = 2;
```
aghaffarkhah: Since we are not using this code in production yet, it is perfectly fine to restart the proto field numbers from 1 again.

dushyantarora: Thanks for the comment. Fixed.

proto/p4/p4runtime.proto
```diff
@@ -300,14 +300,14 @@ message CounterEntry {
...
 // should happen as part of INSERT/DELETE on the associated table-entry.
 // For ReadRequest, the scope is defined as follows:
-// - All counter cells for all meters if counter_id = 0 (default).
-// - All counter cells for given counter_id if table_entry.match is empty.
+// - All counter cells for all tables if table_entry.table_id = 0.
+// - All counter cells of a table if table_entry.table_id is present and
+//   table_entry.match is empty.
 message DirectCounterEntry {
-  uint32 counter_id = 1;
   // The associated table entry match.
```
samar-abdi: suggest: "The associated table entry. This field is required."

antoninbas:
Here is a recap:
- for a WriteRequest: table_entry is required, table_entry.table_id must be valid and the match key must be provided
- for a ReadRequest: table_entry is also always required. A default (0) table_entry.table_id means "read-all entries in all direct counters / meters". Otherwise (valid table_entry.table_id), an empty match means "read-all entries in the direct counter / meter for this specific table" and a valid match key means "read just this specific entry in the direct counter / meter for this specific table".

proto/p4/p4runtime.proto
antoninbas: +1, and same for `DirectMeterEntry`

PR #286, opened 2018-01-23
Support for multiple controller roles

wmohsin:


proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+- Arbitrate that each role has one and only one master client instance
+- Enforce the validity (access control) of P4Runtime requests based on the role config
+
+To keep the protocol simple and extensible, `config` is defined as a `protobuf.Any`,
+and concerete defintions are currently outside the scope of P4Runtime.
+
+### P4Runtime session establishment
+To establish a P4Runtime session, a client needs to open a bi-directional
+stream connection to the server using the `StreamChannel` RPC. The client should advertise
+its `role` and `election_id` right away using a `MasterArbitrationUpdate` message. The process
```
antoninbas: could we mention the case where the `role` is unset here as well (there is a comment in p4runtime.proto but it doesn't appear in the spec)?

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+`PERMISSION_DENIED` will be returned for that update.
 
-Note that for the sake of debugging and testing, we accept `WriteRequest`
+NOTE: All clients are trusted; we do not support the case where a client is using
+an incorrect `role_id` and `election_id` in its `WriteRequest` messages.
+
+NOTE: For the sake of debugging and testing, we accept `WriteRequest`
 messages when there is no master (i.e. no client connections) and the
-`election_id` field is not set.
+`role_id` and `election_id` fields are not set.
```
antoninbas: Apart from the debugging case, I believe we should require `role_id` to be default (0) is this controller instance had an unset `role` field in its `MasterArbitrationUpdate` message.

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
-  for all the slaves.
-
-- otherwise, only the client which sent the request is notified, assuming its
-  election id changed (new connection or election id updated). The `device_id`
-  and `election_id` fields must be set correctly. The `status` field's code is
-  `OK` if that client is the master, `ALREADY_EXISTS` if it is a slave.
-
-### Packet-in
+  - Otherwise, if the `role.id` matches the current one
+    - If the `election_id` matches the current one, the server will accept 'role.config'
```
antoninbas: wrong ticks for role.config

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
 - receive `PacketIn` messages
 - send `PacketOut` messages
 
-### `MasterArbitrationUpdate` client request
+### `MasterArbitrationUpdate` from client to server
 
-- if this is the first `MasterArbitrationUpdate` message sent by the client:
-  - if the election id is already used by another client connection, the server
+- If this is the first `MasterArbitrationUpdate` message sent by the client:
+  - If the election id is already used by another client connection, the server
```
antoninbas: should the election id be unique across all roles or only within a given role id?

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+any given time, only one of the controller instances (for a given role) is considered a master.
+
+### Controller (or P4Runtime *client*) Role
+A given role is identified by `Role.id`.
+The `Role.config` describes the role configuration, i.e. what operations, P4 entities,
+behaviors, etc. are in the scope of a given role. If config is not set (default case),
+it implies that all P4 objects and control behaviors are in scope for that role
+(i.e. full pipeline access).
+
+The Role IDs are defined offline in agreement across the entire control plane.
```
antoninbas: maybe we should decide on "ID" or "id" and use it consistently in non-code sections?

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
-
-### Packet-in
+  - Otherwise, if the `role.id` matches the current one
+    - If the `election_id` matches the current one, the server will accept 'role.config'
+      (if it does not match current one) only if this client is the master.
+    - Otherwise, if the new election id is already used by another client
+      connection, the server should terminate the stream by returning an
+      `INVALID_ARGUMENT` error.
+    - Otherwise, the server updates the `election_id` for this client. If this makes
+      the client the new master, the server will also accept the given 'role.config`.
```
antoninbas: same should be `role.config` instead of 'role.config'

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+  - Otherwise (i.e. `role.id` is different from current one), the server moves the 
+    client to the new role, and processes the update as described above. The server
+    accepts the given `role.config` only if the client becomes master for this
+    role based on the given `election_id`.
+
+### `MasterArbitrationUpdate` from server to client
+
+- If the mastership changed following a `MasterArbitrationUpdate` client
+  request, all existing clients for that role are sent a `MasterArbitrationUpdate`
+  message by the server. The `device_id` must be set. The 'role' and 'election_id'
```
antoninbas: maybe use "`role` and `election_id`" for more consistent formatting

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+  fields must be set to reflect the master client's view. The `status` field code
+  must be `OK` for the master, and `ALREADY_EXISTS` for all the slave clients.
+
+- Otherwise, only the client which sent the request is notified, e.g. a new
+  connection, or a change in `role` and/or `election_id`. The `device_id` must be
+  set. The 'role' and 'election_id' fields must be set to reflect the master
```
antoninbas: same here

proto/docs/arbitration.md
antoninbas: had a discussion with this offline with @wmohsin; seems that we are leaning towards **unique for a given role id**; @wmohsin will update the spec to call this out explicitly

proto/docs/arbitration.md
aghaffarkhah: +1

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+
+P4Runtime supports multiple controller roles to partition the control of the forwarding
+plane and P4 objects. For each role, P4Runtime supports master-arbitration across
+replicated instances of that controller (to protect against multi-master scenarios). At
+any given time, only one of the controller instances (for a given role) is considered a master.
+
+### Controller (or P4Runtime *client*) Role
+A given role is identified by `Role.id`.
+The `Role.config` describes the role configuration, i.e. what operations, P4 entities,
+behaviors, etc. are in the scope of a given role. If config is not set (default case),
```
aghaffarkhah: s/config/role configuration

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
...
+### P4Runtime session establishment
+To establish a P4Runtime session, a client needs to open a bi-directional
+stream connection to the server using the `StreamChannel` RPC. The client should advertise
+its `role` and `election_id` right away using a `MasterArbitrationUpdate` message. The process
+through which election ids are allocated to clients is out-of-scope of P4Runtime and of
+this document. For a  given role, the client with the highest election id is referred to as the
 "master", while all other clients are referred to as "slaves".
 
 Only the master client can successfully:
 - perform `Write` requests
```
aghaffarkhah: Should we use capitals for all the first words in a sentence?

proto/docs/arbitration.md
```diff
@@ -1,74 +1,125 @@
-## Arbitration
```
aghaffarkhah: Hi Waqar can you explain what the expected behavior is when we have an active session with no role config and another session with a given non zero role_id? 

If I receive a stream when there is no role config  (i.e. role_id = 0) I assume this stream will handle all the P4 pipeline entities. Should we explicitly say if you have an active session with no role_id it means the master will be able to handle "all" entities? Or do we want to disallow not giving role_id? I suggest the first for backward compatibility.

proto/docs/arbitration.md
wmohsin: Done

proto/docs/arbitration.md
wmohsin: Called this out in both 'Write RPC' and 'SetForwardingPipelineConfig RPC' sections.

proto/docs/arbitration.md
```diff
@@ -1,74 +1,145 @@
...
+(i.e. full pipeline access).
+
+The Role IDs are defined offline in agreement across the entire control plane.
+A client’s session with P4Runtime is tied to the role ID it passes over the stream
+channel (details below). The P4Runtime server will use role ID for two purposes:
+- Arbitrate that each role has one and only one master client instance
+- Enforce the validity (access control) of P4Runtime requests based on the role config
+
+To keep the protocol simple and extensible, `config` is defined as a `protobuf.Any`,
+and concerete defintions are currently outside the scope of P4Runtime.
```
antoninbas: s/defintions/definitions

wmohsin: Done

PR #292, opened 2018-01-29
Accept unique suffixes as valid names in P4Runtime PTF tests

antoninbas:
p4lang/p4c has been updated to produce correct fully-qualified names
(including the top-level control name) in the P4Info messages. To avoid
modifying existing tests - as well as to provide the ability to use
shorter names in future tests - I have changed the base class for PTF
P4Runtime tests to accept unique suffixes as valid identifiers for
P4Info objects.

### DIFF
--- a/proto/ptf/base_test.py
+++ b/proto/ptf/base_test.py
@@ -18,7 +18,8 @@
 #
 #
 
-from functools import wraps
+from collections import Counter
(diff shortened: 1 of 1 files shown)


antoninbas:
See https://github.com/p4lang/p4c/pull/1106

PR #293, opened 2018-01-30
Require match field to be omitted for don't care

### LINKED ISSUE #35: TableEntry.match semantics for ternary (don't care) match fields
message TableEntry {
  ...
  repeated FieldMatch match = 2;
  ...
}

For a table entry that would like to specify "don't care" values for some of the ternary match fields, we propose the following semantics:
"If a 'match' field has type TERNARY and a value is not provided in a table insert/modify operation, the value will be set to 0 with a 0 mask (i.e. dont care)"

antoninbas:
We have decided to require a match field to be omitted from the
TableEntry for don't care matches. This enables us to achieve smaller
protobuf messages while having a canonical representation for all
matches.

This commit implements the necessary changes to the P4Runtime server,
adds gtests and updates the P4 Runtime specification.

Fixes #35

### DIFF
--- a/proto/docs/p4runtime.md
+++ b/proto/docs/p4runtime.md
@@ -141,12 +141,22 @@ table_entry {
 }
 ```
 
-If `match` for a field is not defined, then the field is treated as don't care
-when constructing the match key for the table entry. Note that leaving match field
-undefined is only allowed for ternary, LPM and range fields. The following table
-entry will program a flow in table `foo` (above) that matches all packets and
-performs action `baz`. Notice the missing fields for `match` and
-`is_default_action`.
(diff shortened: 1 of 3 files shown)


jafingerhut:
You have probably already considered this: Does this change correctly allow the case of a table with only one field that is ternary or lpm, and you want to add a table entry that wildcards all bits in that one field?  Just wanted to double-check there wasn't some gotcha there.

antoninbas:
@jafingerhut yes this case is why we added the `is_default_action` flag:
- empty match key = entry that matches everything
- empty match key + is_default_action set to True = default entry

jafingerhut:
@antoninbas Excellent.  Carry on.

Issue #295, opened 2018-02-02
Reading table entries from table with const entries causes device_mgr.cpp to assert


dushyantarora:
Here is the P4 program:
```
#include <core.p4>
#include <v1model.p4>

struct Headers {}

struct Meta {}

parser p(packet_in b, out Headers h,
         inout Meta m, inout standard_metadata_t sm) {
    state start {
        transition accept;
    }
}

control vrfy(inout Headers h, inout Meta m) { apply {} }
control update(inout Headers h, inout Meta m) { apply {} }

control egress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    apply {}
}

control deparser(packet_out b, in Headers h) {
    apply {}
}

control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    action port_redirect() {
        sm.egress_spec = sm.ingress_port;
    }
    table const_table {
        key = { sm.ingress_port : exact; }
        actions = { port_redirect; }
        const entries = {
          4 : port_redirect();
        }
    }
    apply { const_table.apply(); }
}

V1Switch(p(), vrfy(), ingress(), egress(), update(), deparser()) main;
```

Here is the bmv2 output:
```
New connection
P4Runtime SetForwardingPipelineConfig
P4Runtime Read
device_id: 3
entities {
  table_entry {
  }
}
lt-test_gtest: src/device_mgr.cpp:664: pi::fe::proto::Status pi::fe::proto::DeviceMgrImp::table_read_common(pi::fe::proto::p4_id_t, const pi::fe::proto::common::SessionTemp&, T*, Accessor) const [with T = p4::ReadResponse; Accessor = pi::fe::proto::DeviceMgrImp::table_read_one(pi::fe::proto::p4_id_t, const pi::fe::proto::common::SessionTemp&, p4::ReadResponse*) const::<lambda(p4::ReadResponse*)>; pi::fe::proto::Status = google::rpc::Status; pi::fe::proto::p4_id_t = unsigned int]: Assertion `0 && "Invalid state"' failed.
Aborted (core dumped)
```

antoninbas:
This has been fixed on master, but you will need a recent version of p4c as well.

dushyantarora:
Thanks for fixing this Antonin!

PR #299, opened 2018-02-05
Add is_const_table field to p4info.Table message

antoninbas:
P4 tables with static entries cannot be modified at runtime. It
therefore makes sense to expose that property to the control-plane.

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -116,6 +116,8 @@ message Table {
   repeated uint32 direct_resource_ids = 7;
   int64 size = 8;  // max number of entries in table
   bool with_entry_timeout = 9;  // entry ageing is enabled for table
+  // table with static P4 entries, cannot be modified at runtime
+  bool is_const_table = 10;
 }
 
 // used to list all possible actions in a Table


antoninbas:
@aghaffarkhah @wmohsin @teverman @samar-abdi I'd like to propose this minor change to the P4Info message. While this information may already be present in the config generated by some specific backends, I believe it makes sense to promote it to "first-class citizen" in the P4Info message as the information is useful IMO to the majority of control-planes and P4 Runtime implementations.

proto/p4/config/p4info.proto
```diff
@@ -116,6 +116,8 @@ message Table {
   repeated uint32 direct_resource_ids = 7;
   int64 size = 8;  // max number of entries in table
   bool with_entry_timeout = 9;  // entry ageing is enabled for table
+  // table with static P4 entries, cannot be modified at runtime
```
samar-abdi: LGTM for this change.
What about reading const tables? Should we have a mechanism to filter out const tables while doing a wildcard read from the controller? This would be a useful feature in p4runtime for reconciliation.

antoninbas: I believe a wildcard Read RPC on all tables should return the contents of const tables as well. Maybe the controller can rely on the P4Info message - with this new proposed attribute - to figure out which tables are const and filter out the entries itself?

antoninbas: I'm going to merge this PR so I can make progress on https://github.com/p4lang/PI/issues/295
The issue brought up by @samar-abdi can be discussed at the P4 API WG meeting if time or in a separate Github issue.

Issue #298, opened 2018-02-05
Return ALREADY_EXISTS error code for dup flows in P4 Runtime implementation


antoninbas:
Instead of UNKNOWN

PR #302, opened 2018-02-05
Use correct error codes for TableEntry Write in device mgr

antoninbas:
Use ALREADY_EXISTS for INSERT when entry already exists and NOT_FOUND
for MODIFY / DELETE when entry doesn't exist.


antoninbas:
Fixes #298

antoninbas:
@teverman @dushyantarora thanks for the review!

PR #310, opened 2018-02-13
Added support for Parser Value Set.

samar-abdi:


proto/p4/p4runtime.proto
```diff
@@ -356,6 +357,29 @@ message CloneSessionEntry {
...
 }
 
+//------------------------------------------------------------------------------
+// For writing and reading values in a parser value set. It is OK for the value
+// set to have no entries, as a result of a write. A state transition matching
+// on an empty value set will never be taken. Values contained in a value set
+// must be unique.
+//
+// For Write Requests:
+//    INSERT will add the given values in the repeated field to the existing
```
wmohsin: To be consistent with rest of P4Runtime semantics ... suggest re-aligning this with ValueSet being the P4-entity in this case. So what we insert/modify is the entire value-set, and delete should only require its id.

proto/p4/p4runtime.proto
```diff
@@ -356,6 +357,29 @@ message CloneSessionEntry {
...
+//    DELETE will remove the values given in the repeated field from the value
+//    set if they exist, and will return error if given value(s) are not found.
+//    If only the value_set_id is specified (i.e. values_size() == 0), then all
+//    entries in the value set will be deleted.
+//
+// For Read Requests only the value_set_id should be specified and all the
+// entries in the value set should be returned in the read response.
+message ValueSetEntry {
+  uint32 value_set_id = 1;
+  repeated bytes values = 2;
```
teverman: I went back to the old P4_14 spec to find some background on Value Sets.  I found this statement in section 4.3, "The run time API for updating parser value sets must allow value and mask pairs to be
specified together".  I don't see any place for a mask here, but I don't know whether that is relevant to this revival of value sets.

teverman: In 7.13 of the PSA spec it suggests:
message ValueSetEntry {
        uint32 value_set_id = 1;
        // FieldMatch allows specification of exact, lpm, ternary, and
        // range matching on fields for tables, and these options are
        // permitted for the ValueSet extern as well.
        repeated FieldMatch match = 2;
 }

which is a wider set of values than what this pull request is proposing.  Is this discrepancy a problem.

samar-abdi: Thanks Tom for looking into this. I believe it should be ok to use repeated FieldMatch given @wmohsin suggestion to atomically apply write/read operations across all value set entries.
My only concern is the semantics of the `@size` annotation. Would a range match of 10:20 fit in a value set of size 1? Perhaps @hanw can confirm?

antoninbas: +1 for @teverman's comment on using `FieldMatch` as per the PSA control-plane API proposal
Just like for a table, `size` refers to the logical size, not the actual number of physical entries required in HW.

samar-abdi: BTW, ValueSet will not be a PSA extern (at least in v1.0) but an experimental language feature. It should not have any bearing on the API though.

jafingerhut: @hanw I don't know what the most desirable answer is here, but some thoughts.

If by "logical size" you mean "the number of entries that the control plane wants to add, regardless of whether those entries are arbitrary ranges, exact match, lpm, or ternary entries", then yes, there are hardware implementations such as TCAMs where single range entries in the control plane turn into multiple entries in the hardware, so they are not one-for-one, and not easy to predict in the control plane what they will turn into in a P4 device, and that could differ from device to device, and based upon more criteria than merely whether a range is involved.

This is common between P4 tables and parser value sets.  So nothing really new here.  Whatever answer we want to give for what a 'size' attribute means, I suspect it should be the same for both.

Another thing that might be common to both tables and parser value sets - above a certain minimum size (say a few hundred entries), a table with exact match keys is often most efficiently implemented as a hash table, not a TCAM.  A TCAM implementation gives you nice predictable one-to-one capacity guarantees there between the control plane and data plane.  A hash table does not, not even if you use the common trick of supplementing it with a parallel "overflow TCAM" on the side.  The best you can do is specify some kind of notion of approximate size you want, e.g. I want this exact match table to support at least 800 entries.  A P4 compiler for that target might choose to use a 1024-entry hash table plus a small overflow TCAM for that.  Is it possible that a control plane could try to add 780 entries into this table, and one of them fails?  Unless you have some kind of magic (or you use a hardware TCAM), the answer is "yes, it might fail before you add 800 entries.  Can it be predicted in advance?  Not really, it is dependent on the contents of the particular keys, the hash function(s) in use".

Even more fun, suppose you successfully add 800 entries from the control plane, then reboot your system and the software forgets how those entries were packed into the hardware before.  If the control plane adds the _same_ 800 entries, but in a different order, some of them might fail the next time.

Again, unless you have some magic here, these are just realities of actual implementations, for which the best way to handle them that I know for hash tables are: "put in enough cushion so those failures are rare enough in practice that it doesn't matter for your use case".  Where what constitutes "rare enough" might depend on the use case at hand.

Fun?  Not really.  Reality?  Yes!

proto/p4/p4runtime.proto
samar-abdi: Done.

proto/p4/p4runtime.proto
```diff
@@ -356,6 +357,29 @@ message CloneSessionEntry {
...
+//    INSERT will add the given values in the repeated field to the existing
+//    values in the value set.
+//    MODIFY will replace all existing values in the value set with the given
+//    values in the repeated field.
+//    DELETE will remove the values given in the repeated field from the value
+//    set if they exist, and will return error if given value(s) are not found.
+//    If only the value_set_id is specified (i.e. values_size() == 0), then all
+//    entries in the value set will be deleted.
+//
+// For Read Requests only the value_set_id should be specified and all the
```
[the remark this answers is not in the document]
samar-abdi: Done

proto/p4/p4runtime.proto
samar-abdi: Used logical notion of size (i.e. number of field matches). Seems the simplest to make progress on this.

proto/p4/p4runtime.proto
```diff
@@ -356,6 +357,29 @@ message CloneSessionEntry {
   uint64 packet_length_bytes = 4;
 }
 
+//------------------------------------------------------------------------------
+// For writing and reading matches in a parser value set. A state transition
+// on an empty value set will never be taken. Matches contained in a value set
+// must be unique. The number of matches must be at most the size of the value
```
wmohsin: "must be at most the size of the value set"
nit: Its a little vague where 'size' is coming from ... a reader may think 'size' is part of P4Runtime (e.g. message below). Consider appending "specified in the P4 program".

wmohsin: Regarding uniqueness: Is the expectation that P4Runtime server will validate and flag this by returning an error? (I hope so). If yes, suggest clarifying this (I propose INVALID_ARGUMENT as the code returned for such an update in the WriteRequest).

jafingerhut: Why the requirement for uniqueness?  I can't think of anything that would go wrong in an implementation if it had overlapping ranges, or multiple value/mask entries that both matched the same value.

antoninbas: I like having the uniqueness requirement better. Requiring entries to be unique doesn't mean disallowing overlap between matches. It means rejecting multiple occurrences of the exact same value, just like we did for table entries.

antoninbas: Additionally if we don't add a priority value like we have for table entries, we should probably explicitly call out in the comment that the order of entries in the repeated field determines the priority (with the first value having the highest priority).

samar-abdi: It seems possible to define the same LHS value in the transition select cases for different transitions. So uniqueness is not required. It is kind of trick to implement too. But I do agree with clarifying the priority. Done.

samar-abdi: Done

proto/p4/p4runtime.proto
```diff
@@ -356,6 +357,30 @@ message CloneSessionEntry {
   uint64 packet_length_bytes = 4;
 }
 
+//------------------------------------------------------------------------------
+// For writing and reading matches in a parser value set. A state transition
+// on an empty value set will never be taken. The number of matches must be at
+// most the size of the value set as specified by the @size annotation in P4
+// program. The priority of matches (high to low) is the order in which they
+// have been defined in the repeated field.
```
jafingerhut: I believe that a single value set, no matter how many entries it contains, there is no difference in the parsing behavior regardless of which entry a packet matches.  i.e. it can be considered as an unordered set, with no significance to which entry matches, only whether any of them do, or none of them do.

antoninbas: Correct, that's my bad I misled @samar-abdi :/

proto/p4/p4runtime.proto
```diff
@@ -356,6 +357,29 @@ message CloneSessionEntry {
...
+//    DELETE should only specify the value_set_id and will delete all matches
+//    in the value set.
+//
+// For Read Requests:
+//    All matches for all value-set entries if value_set_id = 0
+//    All matches of the value-set if a valid value_set_id is specified
+//    The 'match' field must never be set in the ReadRequest
+message ValueSetEntry {
+  uint32 value_set_id = 1;
+  repeated FieldMatch match = 2;
```
jafingerhut: FieldMatch should be MatchField?

Bigger question: A table can have one or more fields that it matches on.

A value_set could potentially be allowed to match on multiple fields, too, e.g. if the transition select statement gave a list of fields, like the one in this example P4_16 program: https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/parser-locals2.p4#L76-L80

Does this proposed API enable treating that as a set of "_record_ of field match criteria" (I don't know the right name to use instead of _record_ there -- I simply mean that there could be multiple field match criteria in a single element of the "value set")?

antoninbas: That's a separate question that applies to several other places in P4Runtime and is related to the discussion on type-serialization for control-plane consumption. A value_set should probably not be allowed to match on multiple expressions like a table can - unless there is a use-case. However a single match expression can be a more complex type than just a bitstring, e.g. a tuple of header fields (this applies to both tables & PVS). This question should definitely be answered in some form before the P4Runtime v1.0 release. However we should answer this question generally for the entire P4Runtime API and there is no reason not to move forward with this PR in the mean time.

jafingerhut: Makes sense.  Thanks.

jafingerhut: Actually, rereading your comment confuses me again.  You say "A value_set should probably not be allowed to match on multiple expressions like a table can" and later "a single match expression can be a more complex type than just a bitstring, e.g. a tuple of header fields (this applies to both tables & PVS)".

The latter quote makes sense to me, and the parser value set example I linked to in my earlier comment is exactly that -- a select statement with a tuple of header fields.

How can a table be more general than that, by a "match on multiple expressions"?  Are you referring to the ability of table keys in P4_16 to be expressions like "meta.field_name & 0xff00"?  If not, then I don't understand the distinction you are making.

antoninbas: This is how I see things:
- a table is matching on several separate expressions, each one is a different entry in the "key" property and can have its own static match kind / type. For example:
```
table t {
    key = { expr1 : exact; expr2 : ternary; }
    // ...
}
```
Table `t` matches on 2 expressions and each entry programmed by P4Runtime takes 2 `FieldMatch` values, one for each expression. Most of the time, an expression is just a bitstring (.e.g header field), but we could also match on tuples in a table:
```
table t {
    key = { hdr.ethernet.srcAddr : exact; { hdr.ipv4.srcAddr, hdr.ipv4.dstAddr} : ternary; }
    // ...
}
```
- a parser value set matches on a single expression (which can be of type bitstring, tuple, struct, etc.), with a match kind / type which is dynamic (and not static). Single expression translates into a single `FieldMatch` value in P4Runtime (whereas a table has one per entry in its key).

So in a nutshell I was referring to a table's ability to have several match components in its key (each with its own static match kind), while a PVS has a single match component (with a dynamic match kind).

We will need some work to support complex expressions (such as tuple) in P4Runtime in tables, parser value sets, digest, etc... It is fine IMO to postpone this work for now and merge that proposal as it is. We will need to support these more complex types uniformly in the API. Probably in P4Runtime v1.0.

jafingerhut: FYI, my comments and questions here are _not_ because I am trying to delay this change getting merged as is.  More just trying to understand the plans of what will be possible later.  Thanks.

antoninbas: @jafingerhut my bad, that's not what I meant to imply.
I think serialization of arbitrary P4 types is probably the biggest outstanding question that has to be answered before the v1.0. Maybe that should be the main focus of our next WG meeting?

PR #317, opened 2018-02-20
Port and cos translation in P4Runtime

samar-abdi:


jafingerhut:
Looks good to me.  Glad you have examples of the different P4 program scenarios where these port/cos values can arise.

This would be a perfect place to document (in the future, when we have decided) similar examples for using these types in other externs, e.g. Digest, Register data values, etc.

samar-abdi:
@jafingerhut Thanks for looking into this. I still don't have a great answer on how to deal with ports in PRE entries (multicast groups and clone sessions). Where should these annotations be placed in the P4 code to tell the P4Runtime service that it needs to do translation. The best place would be on the metadata definition in psa.p4. Another (not great) option is to do so on the assignment of the ingress output metadata.

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+  PortIdInHeader_t egress_port;
+}
+```
+The top annotation `@controller_header` is also a standard P4Runtime annotation
+that identifies a header type for a controller packet-out. When the target
+receives a packet-out from the controller over the P4Runtime stream channel, it
+will normally expect packet-out metadata (`egress_port`) value of bitwidth as
+defined by the PSA target's `PortIdInHeader_t` type and in the space of the
+target's port numbers. However, the presence of the `@p4runtime_translation`
+annotation will signal to the p4runtime service that it should expect a
```
teverman: Nit: inconsistent use of "p4runtime service" vs. "P4Runtime service"

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+    drop;
+  }
+}
+```
+Table `t` has an `exact` match on ingress port and `ternary` match on class of
+service metadata. A P4Runtime write request from the controller will have
+the values of the match fields set to controller-specific values. The
+`@p4runtime_translation` annotation on the match fields is a signal to the
+P4Runtime service to intercept these write requests and use the configuration
+data to translate the values to respective device-specific values. Similarly,
```
teverman: I would like to see an expansion of the implications of this translation as they pertain to the "Byte strings" section in the early part of the document.  The PSA spec implies that a type of unspecified width should be treated as 64-bits wide in the P4Runtime API (section 4.1).  Thus, it seems like the P4Runtime service should expect the controller to provide an 8-byte/64-bit string when specifying the match value for istd.ingress_port in table t.  It also seems that the translation can introduce additional errors, depending on how it is implemented and the configuration data provided to the translation logic.  Does the service need to differentiate between pre-translation and post-translation errors?  Is there a difference between a controller-provided string that is a known bad input to the translation logic and another string that fails to translate once it passes as a valid input?

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+annotation will signal to the p4runtime service that it should expect a
+controller defined type and value of `egress_port` metadata. The p4runtime
+service should use the target config to translate this value into PSA
+device-specific value of type `PortIdInHeader_t`.
+
+A similar reverse translation is required in the P4Runtime service for packets
+punted from the target to the controller as shown by the packet-in header
+example below.
+```
+@controller_header("packet_in")
```
antoninbas: This is orthogonal to the PR, but maybe we should consider adding a `p4runtime_` prefix to the `controller_header` annotation, for the sake of consistency?

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+```
+The top annotation `@controller_header` is also a standard P4Runtime annotation
+that identifies a header type for a controller packet-out. When the target
+receives a packet-out from the controller over the P4Runtime stream channel, it
+will normally expect packet-out metadata (`egress_port`) value of bitwidth as
+defined by the PSA target's `PortIdInHeader_t` type and in the space of the
+target's port numbers. However, the presence of the `@p4runtime_translation`
+annotation will signal to the p4runtime service that it should expect a
+controller defined type and value of `egress_port` metadata. The p4runtime
+service should use the target config to translate this value into PSA
```
antoninbas: Nit: "into a PSA device-specific value" sounds better to me

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+The above figure illustrates a motivating example, where an SDN controller is
+controlling multiple P4 targets in a fabric. Switches 1, 2, 3 etc.
+are all targets with different PSA devices, each defining its own port and
+class of service types. In order to uniformly program the fabric, the
+SDN controller may define 32 bit type for ports and 8 bit type for class of
+service. The controller may also define a common number space for ports and
+cannonical classes of service across the fabric. The PSA devices may define
+much shorter bitwidths for these metadata in order to save precious on-chip
+memory in the switch silicon. Furthermore, the values of ports and classes of
+service in the dataplane are PSA device specific and may be different from the
```
antoninbas: Nit: "device-specific"

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+header PacketOut_t {
+  @p4runtime_translation("port")
+  PortIdInHeader_t egress_port;
+}
+```
+The top annotation `@controller_header` is also a standard P4Runtime annotation
+that identifies a header type for a controller packet-out. When the target
+receives a packet-out from the controller over the P4Runtime stream channel, it
+will normally expect packet-out metadata (`egress_port`) value of bitwidth as
+defined by the PSA target's `PortIdInHeader_t` type and in the space of the
```
antoninbas: I thought part of the rationale for having the `InHeader_t` types was that it enabled using the type / bitwidth that the controller actually wanted?
So for this specific case, there will be no type translation, just a value translation, no? (the type "translation" will be a cast in the P4 program, not a software conversion in the agent?)

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+@controller_header("packet_in")
+header PacketIn_t {
+  @p4runtime_translation("port")
+  PortIdInHeader_t ingress_port;
+  @p4runtime_translation("cos")
+  ClassOfServiceInHeader_t cos;
+}
+```
+A packet punted from the target's PSA device will be intercepted by the
+P4runtime service before being sent to the controller. Normally, the P4Runtime
```
antoninbas: Nit: s/P4runtime/P4Runtime

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+service metadata. A P4Runtime write request from the controller will have
+the values of the match fields set to controller-specific values. The
+`@p4runtime_translation` annotation on the match fields is a signal to the
+P4Runtime service to intercept these write requests and use the configuration
+data to translate the values to respective device-specific values. Similarly,
+the P4Runtime service should translate the device-specific values to
+controller-specific values when returning the read response for table `t`.
+
+Note that it may be infeasible to translate the value-mask pair for `ternary`
+matches and the value-prefix-length pair for `lpm` matches. Therefore, the
```
antoninbas: range also?
really, all ternary matches in the general sense of the term

proto/docs/p4runtime.md
```diff
@@ -515,3 +515,164 @@ messages refer to singleton ports by default.
...
+P4Runtime service may require that the match be effectively either exact (all-1
+mask for ternary and prefix-length of the field's controller-bitwidth for lpm)
+or don't care (0 mask for ternary and 0 prefix length for lpm).
+
+### Translation of Action Parameters
+Port and class of service type parameters can be part of a P4 action definition
+as shown in the example below:
+```
+action a(@p4runtime_translation("port") PortId_t p,
```
antoninbas: Looking at this example, I am wondering why the controller-chosen bitwidth is not part of the annotation (e.g. `@p4runtime_translation("port", 32)`. This information would be needed in the P4Info to sanity-check client messages.

In a way this is related to @teverman's question above. Is the controller always sending 64-bit bitstrings? If yes, that doesn't match the example you gave earlier in the doc.

jafingerhut: @antoninbas Are you expecting different controllers to choose different bitwidths for these quantities in the Protobuf messages they send/receive?  If so, for what purpose would they choose different bitwidths?  64 bits should be wide enough for all such occurrences of values of these types, so my first guess is that the motivation would be to economize on bits sent over the wire between client and server?

In any case, I don't see why the bitwidth of these fields between client and server should affect the P4 source code?

proto/docs/p4runtime.md
jafingerhut: My only reason for proposing the InHeader_t types is that they should be a common bit width for _all_ PSA devices.  Yes, if we make them too narrow, then that is bad, which is why I keep asking people to double-check the bit widths chosen at the moment so they are not too small for future uses.

If they are wide enough for all PSA devices, then those widths _could_ also be used as the bitwidths of these fields when present in packet headers for PacketIn / PacketOut packets, and be a common width across all controllers, too.  It isn't necessary, just a possibility.

proto/docs/p4runtime.md
samar-abdi: Done

proto/docs/p4runtime.md
samar-abdi: Agreed.

proto/docs/p4runtime.md
samar-abdi: clarified

proto/docs/p4runtime.md
samar-abdi: done

proto/docs/p4runtime.md
samar-abdi: Done

proto/docs/p4runtime.md
samar-abdi: Done

proto/docs/p4runtime.md
samar-abdi: +1 to @jafingerhut. It is ok for `PortInHeader_t` to be 16 bits and the controller's port type to be 32-bits.

proto/docs/p4runtime.md
samar-abdi: Yes, this was a recent decision. I will update the doc. The controller's port bitwidth is known through the switch config (outside of P4Runtime). It really is not needed in this annotation.

samar-abdi: In P4Info, you really just need a bool to indicate that the field requires translation. I think even that is just a convenience since the annotation will show up anyway. But such a flag will make the switch code more efficient since it won't have to parse the annotations for every field. I can leave that for a different PR if needed.

proto/docs/p4runtime.md
samar-abdi: Great points Tom! Added clarification.

samar-abdi:
It would be nice to get this through for version 1.0 @antoninbas @teverman @wmohsin PTAL.

proto/docs/p4runtime.md
antoninbas: Sorry I just came back from traveling.

I don't believe it is a good idea to introduce that kind of strong asymmetry between regular "fields" and translated "fields". Unless I am mistaken, with your proposal, either there is no more bitwidth information for the field in the p4info or the bitwidth information is meaningless / misleading because it doesn't match what we expect from the controller.

Even if we put aside sanity-checking on the server-side, I believe one should be able to auto-generate a strongly-typed control-plane API based on p4info only, but I don't see how I can do that for translated fields if I'm missing the bitwidth information?

Since this information should be available at compile-time and is not expected to be dynamic, I am not sure why it cannot be included in the annotation & p4info. What's the rationale behind your statement "It really is not needed in this annotation.". Because I can think of situations where it is actually needed or useful to have.

antoninbas: Unless as @jafingerhut wrote, we agree that all translated fields should have bitwidth 64 bit in p4info and be represented as 8-byte bytestrings in p4runtime, which works for me (but I don't see that anywhere in your proposal, even after your most recent changes).

teverman: +1 to Antonin's comments.

I was trying to think of how to express my concern about the indeterminate bitwidth information.

I am also OK with 64 bits and 8-byte strings for the translated fields, but it needs to be documented cleanly.  I really don't like the idea of bitwidths being a config parameter.

samar-abdi: I don't think so. The switch configuration is a configuration of the P4Runtime service as well. We should not be having two different "SDN port bitwidths" defined: one in the config and one in the P4 source. In fact, I am also not very fond of the uint64 types we have for ports/cos in the PRE API. But we can live with them because there is no global field in P4Info that tells us the PSA port/cos P4 bitwidth.

The implementation of P4Runtime service is quite clear:
if annotation: use config bitwidth and translate
else: no translate

Are we worried that the switch may not have a configuration outside P4? If so, a program with such annotations should be rejected by the switch (or the annotations may be ignored).

samar-abdi: Couple of other points:
1. The bitwidth in p4info is only to tell the P4 type (needed if the SDK talks in terms of HW ports only) 
2. The control plane API (I believe you are referring to PD-PI) will know the controller port API also through a similar config that the switch does. And these configs are outside the scope of the P4 program.

My main concern here is that there will be N places for defining SDN types instead on 1. Also, a one-size fits all approach of 64 bit types contradicts a potentially 32-bit type in the config.

samar-abdi: @teverman @antoninbas Can we try to wrap up this PR for Version 1.0? Thanks.

proto/docs/p4runtime.md
```diff
@@ -515,3 +520,176 @@ messages refer to singleton ports by default.
...
+and class of service metadata values between the controller's space and the
+PSA device's space as needed. Such translation is enabled by a standard
+P4Runtime annotation `@p4runtime_translation`. The annotation value is a
+string identifying the metadata type to be translated. Specifically, we define
+* `@p4runtime_translation("port")` for port translation.
+* `@p4runtime_translation("cos")` for class of service translation.
+
+The controller's port/cos bitwidth, the valid SDN port/cos numbers, and their
+mapping to device specific port/cos numbers should be captured in the switch
+config. The P4Runtime service on the target is expected to use the mapping defined
```
jafingerhut: I may be asking a brain-dead question here, but just trying to figure out the intended scenarios you are trying to cover with this, or not cover.

Are you expecting that some controllers will be implemented with a 32-bit port number "compiled into its code", and such controller processes would be _unable_ to communicate with P4Runtime service devices that had a 64-bit port number in their switch config, until and unless one or both sides are changed so that their port bit widths agree with each other?

samar-abdi: I am assuming that all controllers and switches for an SDN domain will agree on a SDN port id type, and use the number space defined by that type to allocate SDN numbers to all ports. This type does not need to be part of the P4 program, because it is already defined in the switch config.

proto/docs/p4runtime.md
antoninbas: Getting this in for v1.0 is definitely the plan, but I don't think we have resolved this discussion yet. Maybe we can discuss it tomorrow at the WG meeting after the type serialization discussion if there is enough time? Otherwise we can always schedule an extra hangout meeting later in the week.

antoninbas:
Superseded by https://github.com/p4lang/p4runtime/pull/50

PR #320, opened 2018-02-24
Wrap indirect resource index in proto message in P4Runtime

### LINKED ISSUE #309: How to do a wildcard Read on an indirect counter / meter?
This is the current p4runtime.proto message to run an entry in an indirect counter?
```
message CounterEntry {
  uint32 counter_id = 1;
  int64 index = 2;
  CounterData data = 3;
}
```
How to disambiguate between reading the counter at index 0 and reading all indices? Usually we leave the field unset to indicate a wildcard Read, but here it doesn't work if 0 is a valid index.

antoninbas:
This ensures that we can distinguish between an unset index (wildcard
read / write) and a zero index (first entry in the resource index).

Fixes #309

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -256,7 +256,7 @@ message Index {
 // For WriteRequest, Update.Type must be MODIFY.
 // For ReadRequest, the scope is defined as follows:
 // - All meter cells for all meters if meter_id = 0 (default).
-// - All meter cells for given meter_id if index = 0 (default).
+// - All meter cells for given meter_id if index is unset (default).
 message MeterEntry {
   uint32 meter_id = 1;
   Index index = 2;
@@ -301,7 +301,7 @@ message MeterConfig {
 // For WriteRequest, Update.Type must be MODIFY.
 // For ReadRequest, the scope is defined as follows:
 // - All counter cells for all meters if counter_id = 0 (default).
-// - All counter cells for given counter_id if index = 0 (default).
+// - All counter cells for given counter_id if index is unset (default).
 message CounterEntry {
   uint32 counter_id = 1;
   Index index = 2;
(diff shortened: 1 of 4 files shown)


proto/p4/p4runtime.proto
```diff
@@ -241,14 +241,25 @@ message ActionProfileGroup {
...
+// message.
+message Index {
+  int64 index = 1;
+}
+
 //------------------------------------------------------------------------------
 // For WriteRequest, Update.Type must be MODIFY.
 // For ReadRequest, the scope is defined as follows:
 // - All meter cells for all meters if meter_id = 0 (default).
 // - All meter cells for given meter_id if index = 0 (default).
```
samar-abdi: If index is omitted from the message

proto/p4/p4runtime.proto
```diff
@@ -293,7 +304,7 @@ message MeterConfig {
 // - All counter cells for given counter_id if index = 0 (default).
```
samar-abdi: if index is omitted from the message

proto/p4/p4runtime.proto
antoninbas: good catch thanks, it is fixed now

proto/p4/p4runtime.proto
```diff
@@ -241,14 +241,25 @@ message ActionProfileGroup {
...
 //------------------------------------------------------------------------------
 // For WriteRequest, Update.Type must be MODIFY.
 // For ReadRequest, the scope is defined as follows:
 // - All meter cells for all meters if meter_id = 0 (default).
-// - All meter cells for given meter_id if index = 0 (default).
+// - All meter cells for given meter_id if index is unset (default).
 message MeterEntry {
   uint32 meter_id = 1;
-  int64 index = 2;
+  Index index = 2;
```
teverman: Technically, if the type changes, it can cause compatibility issues, and a new field number should be used.  We should be OK on this one because I don't see any references to either existing index outside of p4lang/PI.

antoninbas: agreed, but I figured few people outside of myself must be using this, and we haven't technically released a stable P4Runtime version yet

PR #328, opened 2018-03-23
Support for arbitrary type serialization in P4Runtime

antoninbas:
Adds support for arbitrary P4_16 types in P4Runtime, as discussed at the
03/21/2018 WG meeting (refer to minutes).

The changes incur minimal overhead for the most-common case (`bit<W>` in
P4, for which we use a binary string in P4Runtime), and enable us to
support complex types in lookup expressions and PSA externs such as
Register and Digest.

We include some documentation for the type specifications (P4Info) and
value representations (P4Runtime) which can be merged in the main
P4Info/P4Runtime specification in the future (before the v1.0 release).

To avoid breaking existing implementations, we do not use the new P4Data
representation for tables and value-sets.

### DIFF
--- a/proto/docs/p4runtime_type_serialization.md
+++ b/proto/docs/p4runtime_type_serialization.md
@@ -1,0 +1,96 @@
+# Arbitary P4_16 type serialization in P4Runtime
+
+## Problem statement
+The P4_16 language includes more complex types than just bitstrings or even
+headers (see [P4_16
+specification](https://p4.org/p4-spec/docs/P4-16-v1.0.0-spec.html#sec-p4-type)). Most
+of these complex data types can be exposed to the control-plane through table
+key expressions, value set lookup expressions, register (PSA extern type) value
+types, etc. Not supporting these more compex types can be very limiting. For
+example, the following P4_16 objects involve complex types that need to be
+exposed in P4Runtime in order to support runtime operations on these objects.
+
+```
+value_set<tuple<bit<16>, bit<8> > > pvs_complex;
+state parse_ipv4 {
+   packet.extract<ipv4_t>(hdr.ipv4);
+   transition select({ hdr.ipv4.version, hdr.ipv4.protocol }){
+       pvs_complex: parse_inner;
+       default: accept;
+   }
+}
+```
+
+```
+header_union ip_t {
+   ipv4_t ipv4;
+   ipv6_t ipv6;
+}
+Register<ip_t, bit<32> >(128) register_ip;
+```
+
+One solution would be to use only bitstrings (`bytes` type) in p4runtime.proto
+and to define a custom serialization format for complex P4_16 types. The
+serialization would maybe be trivial for header types but would require some
+work for header unions, stacks, etc. Rather than coming-up with a serialization
+format from scratch, we decided to use a Protobuf representation for all P4_16
+types, since the P4Runtime service already uses Protobuf messages.
+
+## P4 type specifications in p4info.proto
+In order for the P4Runtime client to generate correctly-formatted messages and
+for the P4Runtime service implementation to validate them, P4Info needs to
+specify the type of each P4 expression which is exposed to the control-plane. In
+the register example above, client and server need to know that each element of
+the register has type `ip_t`, which is a header union with 2 possible headers:
+`ipv4` with type `ipv4_t` and `ipv6` with type `ipv6`. Similarly, they need to
+know the field layout of both these header types.
+
+To achieve this we introduce 2 main protobuf messages: `P4TypeInfo` and
+`P4DataTypeSpec`.
+
+`P4TypeInfo` is a top-level member of `P4Info` and includes Protobuf maps
+storing the type specification for all the named types in the P4_16
+program. These named types are `struct`, `header`, `header_union` and `enum`;
+for each of these we have a type specification message, respectively
+`P4StructTypeSpec`, `P4HeaderTypeSpec`, `P4HeaderUnionTypeSpec` and
+`P4EnumTypeSpec`. We support P4 annotations for named types, which is useful to
+identify well-known headers, such as IPv4 or IPv6. `P4TypeInfo` also include the
+list of parser errors for the program, as a `P4ErrorTypeSpec` message.
+
+`P4DataTypeSpec` is meant to be used in `P4Info`, everywhere where the P4Runtime
+client can provide a value for a P4_16 expression. `P4DataTypeSpec` describes
+the compile-time of the expression as a Protobuf `oneof`, which is a `string` in
+case of a named type (`struct`, `header`, `header_union` or `enum`), an empty
+Protobuf message for `bool` and `error`, and a Protobuf message for other
+anonymous types (`bit<W>`, `int<W>`, `varbit<W>`, `tuple` or `stack`). The
+"bitstring" types (`bit<W>`, `int<W>` and `varbit<W>`) are grouped together in
+the `P4BitstringLikeTypeSpec` message, since they are the only sub-types allowed
+in headers and values with one of these types are represented similarly in
+P4Runtime (with binary strings).
+
+## P4 data in p4runtime.proto
+P4Runtime uses the `P4Data` message to represent values with arbitrary
+types. The P4Runtime client must generate correct `P4Data` messages based on the
+type specification information included in `P4Info`. The `P4Data` message was
+designed to introduce little overhead compared to using binary strings in the
+most common case (P4_16 `bit<W>` type).
+
+Just like its `P4Info` counterpart `P4DataTypeSpec`, `P4Data` uses a `oneof` to
+represent all possible values.
+
+### `enum` and `error`
+We currently use human-readable `string` in `P4Data` to represent `enum` and
+`error` values. Indeed, the current P4_16 specification does not specify any
+mechanism through which integer values are assigned to `enum` and `error`
+members -whether automatically by the compiler or by letting the programmer pick
+values. We may switch to an `int32` representation if the P4 language is updated
+to specify the underlying representation of enums, in which case we would also
+include a mapping from name to integer value in the `P4TypeInfo` message.
+
+## Trade-off for v1.0 release
+For the v1.0 release of P4Runtime, the Working Group has decided not to replace
+occurences of `bytes` with `P4Data` in the `FieldMatch` message, which is used
+to represent table and value set entries. This is to avoid breaking
+already-existing implementations of P4Runtime. However `P4Data` is used whenever
+appropriate for PSA externs and we encourage the use of `P4Data` in
+architecture-specific extensions.
--- a/proto/docs/p4runtime_type_serialization.md
+++ b/proto/docs/p4runtime_type_serialization.md
@@ -72,7 +72,7 @@ For all P4_16 compound types (`tuple`, `struct`, `header`, and `header_union`),
 the order of members in the `repeated` field of the Protobuf type specification
 is guaranteed to be the same as the order of the members in the corresponding
 P4_16 declaration. The same goes for the order of members of an `enum` or
-members of `error`, as well as for the order of entries is a `stack`.
+members of `error`, as well as for the order of entries in a `stack`.
 
 ## P4 data in p4runtime.proto
 P4Runtime uses the `P4Data` message to represent values with arbitrary
--- a/proto/docs/p4runtime_type_serialization.md
+++ b/proto/docs/p4runtime_type_serialization.md
@@ -68,6 +68,12 @@ the `P4BitstringLikeTypeSpec` message, since they are the only sub-types allowed
 in headers and values with one of these types are represented similarly in
 P4Runtime (with binary strings).
 
+For all P4_16 compound types (`tuple`, `struct`, `header`, and `header_union`),
+the order of members in the `repeated` field of the Protobuf type specification
+is guaranteed to be the same as the order of the members in the corresponding
+P4_16 declaration. The same goes for the order of members of an `enum` or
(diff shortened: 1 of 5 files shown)


antoninbas:
@jafingerhut for some reason I cannot add you as a reviewer but it would be great if you could take a look at this
@mbudiu-vmw it would be great if you take a quick look and point-out if you think I missed anything

proto/docs/p4runtime_type_serialization.md
```diff
@@ -0,0 +1,96 @@
...
+
+To achieve this we introduce 2 main protobuf messages: `P4TypeInfo` and
+`P4DataTypeSpec`.
+
+`P4TypeInfo` is a top-level member of `P4Info` and includes Protobuf maps
+storing the type specification for all the named types in the P4_16
+program. These named types are `struct`, `header`, `header_union` and `enum`;
+for each of these we have a type specification message, respectively
+`P4StructTypeSpec`, `P4HeaderTypeSpec`, `P4HeaderUnionTypeSpec` and
+`P4EnumTypeSpec`. We support P4 annotations for named types, which is useful to
```
mbudiu-vmw: I wonder whether you will care about serializing match_kind ever.
It's just a special enum.

proto/docs/p4runtime_type_serialization.md
```diff
@@ -0,0 +1,96 @@
...
+types. The P4Runtime client must generate correct `P4Data` messages based on the
+type specification information included in `P4Info`. The `P4Data` message was
+designed to introduce little overhead compared to using binary strings in the
+most common case (P4_16 `bit<W>` type).
+
+Just like its `P4Info` counterpart `P4DataTypeSpec`, `P4Data` uses a `oneof` to
+represent all possible values.
+
+### `enum` and `error`
+We currently use human-readable `string` in `P4Data` to represent `enum` and
```
mbudiu-vmw: People want the ability to specify this in the P4 program - perhaps using an annotation.
It would make sense to allow this - sometimes wasting 32 bits on enums may be too expensive.

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+  map<string, P4EnumTypeSpec> enums = 4;
+  P4ErrorTypeSpec error = 5;
+}
+
+// Describes a P4_16 type.
+message P4DataTypeSpec {
+  oneof type_spec {
+    P4BitstringLikeTypeSpec bitstring = 1;
+    P4BoolType bool = 2;
+    // From the P4_16 spec: "A tuple is similar to a struct, in that it holds
```
teverman: Nit: I'd prefer to see this comment right before message P4TupleSpec

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+ * | struct             | error     | error        | allowed         |
+ * | tuple              | error     | error        | allowed         |
+ * |--------------------|-----------|--------------|-----------------|
+ */
+
+// These P4 types (struct, header_type, header_union and enum) are guaranteed to
+// be "named" (e.g. you cannot use an anonymous struct to declare a variable,
+// like in C). Instead of duplicating the type spec for these every time the
+// type is used, we include the type spec once in this P4TypeInfo message and
+// refer to the types by name in the P4DataTypeSpec message. We also support
```
teverman: Is "name" always the fully-qualified name?

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+
+message P4ErrorType { }
+
+message P4BitstringLikeTypeSpec {
+  oneof type_spec {
+    P4BitTypeSpec bit = 1;  // bit<W>
+    P4IntTypeSpec int = 2;  // int<W>
+    P4VarbitTypeSpec varbit = 3;  // varbit<W>
+  }
+  // Useful to identify well-know types, such as IP address or Ethernet MAC
```
teverman: know->known

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+  repeated Member members = 1;
+  repeated string annotations = 2;
+}
+
+message P4HeaderTypeSpec {
+  message Member {
+    string name = 1;
+    P4BitstringLikeTypeSpec type_spec = 2;
+  }
+  repeated Member members = 1;
```
teverman: What needs to be said about the order of members?  Can an implementation expect that for any valid value of N, the offset of members(N) in the header is greater than the offset of members(N-1) from the start of the header?

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+}
+
+message P4HeaderUnionStackTypeSpec {
+  P4NamedType header_union = 1;
+  int32 size = 2;
+}
+
+// As of today, P4_16 enum members have no integer value assigned to
+// them. Future versions of this protobuf interface may use integer values for
+// enums in P4Data, which means the repeated field would be replaced by a map in
```
teverman: I'm contemplating a suggestion to make it a map now to make it future-proof, but it seems like that may be excess complexity if integer values never come to pass.

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+// this message (mapping from member names to integer values). This is dependent
+// on future versions of the P4 language.
+message P4EnumTypeSpec {
+  repeated string members = 1;
```
teverman: What about the ability to annotate individual members of the enum?  Does the language allow it?

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+  repeated string annotations = 2;
+}
+
+// Similar to an enum, but there is always one and only one instance per P4
+// program.
+message P4ErrorTypeSpec {
+  repeated string members = 1;
```
teverman: Are annotations relevant?

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// End of P4 type specs --------------------------------------------------------
+
+message P4Data {
```
teverman: Will it be possible to define pass an action parameter as P4Data?

proto/p4/p4types.proto
```diff
@@ -0,0 +1,207 @@
...
+  oneof data {
+    bytes bitstring = 1;  // for bit<W>, int<W>, varbit<W>
+    bool bool = 2;
+    P4StructLike tuple = 3;
+    P4StructLike struct = 4;
+    P4Header header = 5;
+    P4HeaderUnion header_union = 6;
+    P4HeaderStack header_stack = 7;
+    P4HeaderUnionStack header_union_stack = 8;
+    // Could be replaced with integer values in future versions.
```
teverman: I suppose that in the future, some programmers may choose to assign integer values to some subset of enums, while others may choose to retain named enums, so wouldn't this need to allow for both?

proto/p4/p4types.proto
antoninbas: done

proto/p4/p4types.proto
antoninbas: clarified by using "fully-qualified name" in comment

proto/p4/p4types.proto
antoninbas: done

proto/p4/p4types.proto
antoninbas: I just updated the documentation to specify that the order of the repeated "members" field for compound types must match the order in the corresponding P4 declaration. I didn't add comments to the proto, since I would need to have the same comment in a dozen place.

proto/p4/p4types.proto
antoninbas: I think that we should stick with it until there is progress on the language side. We can then deprecate the field and introduce new ones if it makes sense.

proto/p4/p4types.proto
antoninbas: My reasoning was that I would not try to anticipate where annotations could be useful. So I only put annotations where I knew for sure they could be useful. Maybe we can plan on adding the annotations field where needed when the need arises? This can be done easily in a backward-compatible way.

proto/p4/p4types.proto
antoninbas: Good point. Yes in theory. However for the time being I imagine we would like to stick with `bytes` (just like for lookup fields). So I updated the spec to that effect.

proto/p4/p4types.proto
antoninbas: I believe that before the programmer is allowed to choose integers, the language will need to describe how integer values are assigned **by default**, i.e. with programmer intervention. Which means that even when the programmer doesn't provide a specific value, the compiler assigns one. If this change happens in the P4 language, we can safely deprecate the `string` field and add `int32` or `int64` fields here, with a mapping in P4Info (name -> int), since all enum members will always have a integer value.

proto/docs/p4runtime_type_serialization.md
antoninbas: I cannot think of a case where that would be useful to the control plane. From a language perspective, can I even instantiate a variable of type `match_kind`?

antoninbas:
@chrispsommers this PR may also be of interest to you

proto/docs/p4runtime_type_serialization.md
```diff
@@ -0,0 +1,109 @@
...
+"bitstring" types (`bit<W>`, `int<W>` and `varbit<W>`) are grouped together in
+the `P4BitstringLikeTypeSpec` message, since they are the only sub-types allowed
+in headers and values with one of these types are represented similarly in
+P4Runtime (with binary strings).
+
+For all P4_16 compound types (`tuple`, `struct`, `header`, and `header_union`),
+the order of members in the `repeated` field of the Protobuf type specification
+is guaranteed to be the same as the order of the members in the corresponding
+P4_16 declaration. The same goes for the order of members of an `enum` or
+members of `error`, as well as for the order of entries is a `stack`.
```
jafingerhut: typo: is -> in ?

jafingerhut:
I have read through the changes, and except for a minor typo I cannot think of any comments or questions that other reviewers have not already made.  Thanks.

proto/p4/p4types.proto
jafingerhut: I can imagine the P4_16 language spec allowing some enum's to have _no_ definition of integer values for enum members, in a case where they never need to communicate that enum value into or out of the P4 program to other places (whether that other place is the control plane, or another data plane device).

Whether there is enough of an advantage for the language to require the programmer to specify integer values for all enum members for those enum types that need them, and leave out integer value specifications by the programmer for other enum types that do not need them, I can raise that question in the LDWG when it comes up.  I don't have a strong opinion either way on that detail.

I have a mild opinion that having some enums with integer values specified, and other enums without the integer values specified, adds extra error checks and 2 cases to think about in lots of places involving enum's, instead of only 1 case.

proto/p4/p4types.proto
chrispsommers: Per the P4_16 spec section 7.2.1, enums can be annotated:
```
enumDeclaration
    : optAnnotations ENUM name '{' identifierList '}'
```
At some point I think carrying these annotations into P4Info would be valuable.  I have several ideas related to automatic user-interface generation where annotating an enum (and specifying a vaue for  it)  could be nice.

antoninbas: Thanks for the feedback. My proposal already supports annotations on enums, just not on enum *members* (not that the language syntax does not support these either). I wanted to wait until there was interest before adding support for annotations on enum members. However, since both you and Tom have brought-up this point, I have updated my proposal to support annotations on enum members. It will indeed be useful to assign integer values to enum members until the LGDW standardizes something.

chrispsommers: @antoninbas It looks like https://github.com/p4lang/PI/commit/7cb4fb5de63a4c39c4e1dfc6b7fd4e92a81f4dbe reverted your change?

chrispsommers: You're right, it looks OK in the final PR. I got lost in the rabbit hole. Thanks!

chrispsommers: @antoninbas I can't seem to confirm that P4_16 accepts annotations on enum _members_ or more generally identifiers in an identifier list.
```
enumDeclaration
    : optAnnotations ENUM name '{' identifierList '}'
...
identifierList
: name
| identifierList ',' name
;
```

antoninbas:
Thanks to everyone who reviewed. This is the last call for comments, I will be merging this on Monday.

jafingerhut:
My comment here should not block this merge from happening, but I wanted to ask to see if it might result in a future change to this later.

If the idea of handling PortId_t by defining it as `struct PortId_t { bit<N> value; }` becomes accepted, then would that mean that using type PortId_t value as fields in a table key, or an action parameter, would need to handle these types in the P4-API?  That _could_ be done by using the whole general type serialization approach for all table keys and action parameters, but I have a rough memory that in the P4-API working group, some people wanted to delay that for after v1.0, and restrict table key fields and action parameters to `bit<W>` types for v1.0?

antoninbas:
@jafingerhut As I hinted to in my comment there (https://github.com/p4lang/p4c/issues/1157#issuecomment-377331736), I think the best thing to do would be to use a 32-bit binary string for PortId_t in P4Info / P4Runtime - which basically amounts to treating PortId_t and other translated values in a special way. The P4Info serializer in the compiler would return an error if a table matches on a struct which is not one of PortId_t, ClassOfService_t, etc...

jafingerhut:
@antoninbas Thanks for explanation.  Making special cases for type names PortId_t, ClassOfService_t, and any of a tiny number of others seems like a pragmatic approach for now to get things moving.

PR #329, opened 2018-03-27
Add support for P4Runtime RECONCILE_AND_COMMIT

antoninbas:
As an action mode for SetForwardingPipelineConfig in reference
implementation.

We use a pretty naive implementation: we read all entries using a
wildcard read, perform the update sequence then write all entries back.

This commit is also the perfect opportunity to introduce proper
top-level locking to synchronize all RPCs. This introduces a dependency
on the Boost thread library as we want to use a shared mutex (not
available in the C++11 standard library). We allow concurrent Write
calls; however Read and SetForwardingPipelineConfig get an exclusive
lock. These Read / Write semantics are as agreed on for P4Runtime by the
P4 API WG.

### DIFF
--- a/CLI/table_dump.c
+++ b/CLI/table_dump.c
@@ -138,8 +138,8 @@ static pi_cli_status_t dump_entries(pi_p4_id_t t_id,
     }
 
     uint32_t priority = pi_match_key_get_priority(entry.match_key);
-    // TODO(antonin): bmv2 quirk, for tables with no priority, -1 is returned
-    if (priority != (uint32_t)-1) printf("Priority: %u\n", priority);
+    // TODO(antonin): 0 means no priority?
+    if (priority != 0) printf("Priority: %u\n", priority);
 
     print_action_entry(&entry.entry);
   }
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,10 +20,12 @@ ENV PI_DEPS automake \
             g++ \
             libboost-dev \
             libboost-system-dev \
+            libboost-thread-dev \
             libtool \
             pkg-config \
             libjudy-dev
 ENV PI_RUNTIME_DEPS libboost-system1.58.0 \
+                    libboost-thread1.58.0 \
                     libjudydebian1 \
                     python
 
--- a/Dockerfile.bmv2
+++ b/Dockerfile.bmv2
@@ -29,6 +29,7 @@ ENV PI_DEPS automake \
             g++-6 \
             libboost-dev \
             libboost-system-dev \
+            libboost-thread-dev \
             libtool \
             libtool-bin \
             pkg-config \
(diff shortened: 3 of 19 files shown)


jafingerhut:
Some implementations will have the possibility of the following sequence of events, I think:

+ reading all existing table entries, and SetForwardingPipelineConfig executes, all completing successfully
+ while writing the entries back into the tables, a failure occurs (e.g. because the table is implemented via a hash table where success/failure depends upon the order of operations, and these writes are in a different order than the ones done to get the table into this state earlier).

In this case, will this API call return some kind of error status to the controller?

antoninbas:
That's a good point Andy. At the moment, if the write operation (that replays all the flows) fails, a (not very helpful) INTERNAL error will be returned to the controller.

I implemented this feature so that people can start using it on bmv2, which will not have this problem. This implementation is probably very inadequate for many targets, in particular fixed-function or semi fixed-function ASICs.

I think that in the case of RECONCILE_AND_COMMIT a lot of the burden is on the controller, which needs to know when it makes sense to use it and has to be able to recover in case of failure. I believe that one valid use case for RECONCILE_AND_COMMIT is when table sizes are modified in the P4. If a table is shrunk, it is up to the controller to make sure that the existing table doesn't have too many flows for the new size. As you pointed out when discussing the "size" property for P4_16, there is no easy way to guarantee that for hash tables but assuming reasonable occupancy (or if the compiler over-provisions) the probability of failure is small.

On a related note, I plan to implement the necessary logic to detect if the P4 configuration hasn't changed at all, in which case RECONCILE_AND_COMMIT will be a no-op (unless I'm missing something?).

jafingerhut:
I am not aware of anything you are missing in your last question.

One more I thought of: Does RECONCILE_AND_COMMIT bring down ports, and cause them to remain down until after all table entries are added successfully?  The fact that they are being added in an order that is arbitrary relative to the unknown behavior of the P4 program could cause the forwarding state of the device to go through many intermediate states that the controller might not want packets flowing through during that time.

antoninbas:
I'm not sure how easy it will be to standardize such a thing but I think I would be happy with something along of the lines of: "If the RECONCILE_AND_COMMIT causes a change in the hardware forwarding state of the device, the device must stop accepting traffic (i.e. drop packets) until the RECONCILE_AND_COMMIT completes. If the RECONCILE_AND_COMMIT fails for any reasons, the device will not start accepting traffic again until the client takes the appropriate action (e.g. a VERIFY_AND_COMMIT operation). If the RECONCILE_AND_COMMIT fails, the client may not be able to perform a new RECONCILE_AND_COMMIT immediately as the switch forwarding state may be lost; the client will need to replay the forwarding state."

I wouldn't talk about ports at all. In the case of a fixed-function device, I would imagine that most RECONCILE_AND_COMMIT operations could be done without the ports flapping and without any traffic loss.

PR #336, opened 2018-04-05
Add definition of id prefixes to p4info.proto as enum

antoninbas:
Historically ids have shared the same number-space (i.e. all ids for
top-level objects are unique). Furthermore, the compiler has been
allocating ids in such a way that it is possible to deduce the object
types (e.g. table, action, ...) based on the value of the
most-significant byte of the id.

We decided that it was a bad idea to change that scheme and break
existing implementations of P4Runtime. However, it was also decided that
we should add an enum to p4info.proto so that these 1-byte values are
documented in code somewhere. The compiler should use these values when
allocating ids for the P4 objects.

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -47,58 +47,58 @@ message Documentation {
 // Top-level package documentation describing the forwarding pipeline config
 // Can be used to manage multiple P4 packages.
 message PkgInfo {
-    // a definitive name for this configuration, e.g. switch.p4_v1.0
-    string name = 1;
-    // configuration version, free-format string
-    string version = 2;
-    // brief and detailed descriptions
-    Documentation doc = 3;
-    // Miscellaneous metadata, free-form; a way to extend PkgInfo
-    repeated string annotations = 4;
-    // the target architecture, e.g. "psa"
-    string arch = 5;
-    // organization which produced the configuration, e.g. "p4.org"
-    string organization = 6;
-    // contact info for support,e.g. "tech-support@acme.org"
-    string contact = 7;
-    // url for more information, e.g. "http://support.p4.org/ref/p4/switch.p4_v1.0"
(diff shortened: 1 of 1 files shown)


chrispsommers:
Good idea, it needs to be nailed down and formalized.

proto/p4/config/p4info.proto
```diff
@@ -65,6 +65,40 @@ message PkgInfo {
...
+
+        // P4 language built-ins
+        ACTION = 0x01;
+        TABLE = 0x02;
+        VALUE_SET = 0x03;
+
+        // PSA externs
+        PSA_EXTERNS_START = 0x10;
+        ACTION_PROFILE = 0x11;
+        COUNTER = 0x12;
```
samar-abdi: Are we differentiating between direct and indirect resources?

proto/p4/config/p4info.proto
```diff
@@ -65,6 +65,40 @@ message PkgInfo {
...
+        METER = 0x13;
+
+        // externs for other architectures (vendor extensions)
+        OTHER_EXTERNS_START = 0x80;
+
+        // max value for an unsigned 8-bit byte
+        MAX = 0xff;
+        // requires protoc >= 3.5.0
+        // reserved 0x100 to max;
```
teverman: This comment is a little misleading.  Any value 0x100 and up is invalid since this enum defines the most significant byte of an ID.  "Reserved" implies they could be used someday.

antoninbas: Isn't the "reserved" keyword a good way to ensure that we will not use these values in the future? That being said, I don't really mind removing it as "reserved" requires a very recent version of protoc.

teverman: My brain locked in to the comment // preceding the reserved keyword, and it didn't even register as a protoc statement.

PR #344, opened 2018-04-20
Add missing id prefixes in p4info.proto

antoninbas:
For PSA externs Register and Digest, as well as for controller packet
headers (header types annotated with @controller_header in P4).

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -83,6 +83,7 @@ message P4Ids {
     ACTION = 0x01;
     TABLE = 0x02;
     VALUE_SET = 0x03;
+    CONTROLLER_HEADER = 0x04;
 
     // PSA externs
     PSA_EXTERNS_START = 0x10;
@@ -91,6 +92,8 @@ message P4Ids {
     DIRECT_COUNTER = 0x13;
     METER = 0x14;
     DIRECT_METER = 0x15;
+    REGISTER = 0x16;
+    DIGEST = 0x17;
 
     // externs for other architectures (vendor extensions)
     OTHER_EXTERNS_START = 0x80;


proto/p4/config/p4info.proto
```diff
@@ -91,6 +92,8 @@ message P4Ids {
     DIRECT_COUNTER = 0x13;
     METER = 0x14;
     DIRECT_METER = 0x15;
+    REGISTER = 0x16;
+    DIGEST = 0x17;
```
jafingerhut: Should ACTION_SELECTOR be added to this list as well?  Either that or you have chosen to "merge" it with ACTION_PROFILE, which is already there.

Also, is PSA_EXTERNS_START = 0x10 a "dead value" here, i.e. would the numeric value 0x10 ever be used for any PSA extern objects?  It seems a shame to waste a whole 8-bit value a "gap" between two consecutive ranges like this.  I don't know if Protobuf syntax would let you assign both the names PSA_EXTERNS_START and ACTION_SELECTOR to the value 0x10?

proto/p4/config/p4info.proto
```diff
@@ -83,6 +83,7 @@ message P4Ids {
     ACTION = 0x01;
     TABLE = 0x02;
     VALUE_SET = 0x03;
+    CONTROLLER_HEADER = 0x04;
```
jafingerhut: What in a P4 program would become CONTROLLER_HEADER?  Sorry, I may be behind on the thinking here.

proto/p4/config/p4info.proto
antoninbas: action_selector is merged with action_profile
Ideally, we would have "one psa extern" <-> "one p4runtime entity" <-> "one id prefix value". However, where this falls short, I believe we should favor "one p4runtime entity" <-> "one id prefix value". Since we are unlikely to modify P4Runtime for action profiles / selectors (this is already actively used), that's what I'm going with here.

I agree that it's a shame to waste 4 values in total here. But I think we are unlikely to run out of values for P4_16 built-in objects & PSA externs. If a target needs more than 120 extern types (which I find quite unlikely), a couple extra values is probably not going to make much difference.

proto/p4/config/p4info.proto
antoninbas: Header types annotated with `@controller_header` (this is referred to as `controller_packet_metadata` in the P4Info). This is not architecture-specific IMO, so I added it next to the P4 built-in objects.

PR #346, opened 2018-04-24
Add support for idle timeout for table entries in P4Runtime

### LINKED ISSUE #265: Add idle timeout support to P4 Runtime `TableEntry` message
We need to discuss how we plan on supporting table entry timeout in P4 Runtime. I'm assuming that there is interest in supporting this since this is a common OpenFlow feature (e.g. see flow entry timeouts in https://www.opennetworking.org/images/stories/downloads/sdn-resources/onf-specifications/openflow/openflow-spec-v1.4.0.pdf) and AFAIK is also a common feature in various switch SDKs.

This issue was brought up at the 12/20/2017 PSA WG meeting during a discussion regarding the need for a `idle_timeout` table property to indicate that the match-table should support entry eviction. Minutes for the PSA WG meeting can be found here: https://github.com/p4lang/p4-spec/wiki/PSA-Meeting-Minutes-Dec-20,-2017. At the meeting it was agreed that the table property format would depend on the level of support for idle timeout in P4 Runtime.

As a result, I would like to propose 3 different solutions.
For all these proposals, we add a `uint64 timeout_ns` field to the `TableEntry` protobuf message (https://github.com/p4lang/PI/blob/master/proto/p4/p4runtime.proto#L119) to indicate the timeout in nanoseconds for the entry. We can decide that a value of `0` indicates that the entry is never to be evicted.

Terms I use in the proposal:
  - hard-eviction of entries: the P4 Runtime implementation is in charge of deleting the entry when the timeout expires. This can be done in hardware (for targets which support it) or in the local software stack. Each target is free to implement the feature in any way it wants.
  - streaming of notifications to control-plane: when an entry timeout expires, we generate a P4 Runtime notification (message format TBD) which is streamed to the controller / NOS. The control-plane (local or remote) is in charge of explicitly deleting the entry with a `Write` RPC. The P4 Runtime implementation does not evict table entries itself.

## Proposal 1: boolean table property with hard eviction of entries
With this proposal we define a `psa_idle_timeout` table property which can take values `True|False`. The P4 Runtime implementation is responsible for deleting entries on its own when the timeout expires. We may decide to generate a P4 Runtime notification ex post facto to the control-plane to indicate that an entry has been deleted.

## Proposal 2: boolean table property with streaming of notifications to control-plane
With this proposal we define the same `psa_idle_timeout` table property as for the first proposal. The P4 Runtime implementation must not evict entries of its own accord. Instead it generates P4 Runtime notifications (similar to digest notifications: https://github.com/p4lang/PI/pull/242) which the control-plane can then act on.

## Proposal 3: enum table property with either hard eviction or streaming of notifications
This is the most flexible proposal. We define the following `enum` in psa.p4:
```
enum IdleTimeoutBehavior {
    NO_TIMEOUT,  // default if psa_idle_timeout property missing
    HARD_EVICTION,
    NOTIFIY_CONTROL_ONLY
}
```
The table property `psa_idle_timeout` can then use any member of the enum (which member is used will be reflected in the `P4Info` message). P4 Runtime will accommodate for all possibilities.
We can even imagine more choices:
```
enum IdleTimeoutBehavior {
    NO_TIMEOUT,  // default if psa_idle_timeout property missing
    HARD_EVICTION_WITH_NOTIFICATION,  // notification is sent to control after evicting
    HARD_EVICTION_WITHOUT_NOTIFICATION,  // silently evict
    NOTIFIY_CONTROL_ONLY
}
```

antoninbas:
Necessary changes to p4info.proto and p4runtime.proto to support idle
timeout for table entries. A table supports idle timeout if the
idle_timeout_behavior field in the corresponding P4Info Table message is
set to NOTIFY_CONTROL (we assume that this information comes from the P4
table definition, either as a table attribute or as an annotation).

When a table supports idle timeout, the P4Runtime client can specifiy a
TTL for each TableEntry. This TTL value will be returned to the client
during a Read operation. If an entry hasn't been hit in the last TTL
nanoseconds, a notification is sent to the client on the
StreamChannel. After a notification is sent, the P4Runtime server resets
the timer for this entry, which means a new notification will be sent
after another TTL nanoseconds if the entry is not hit. This means that
it is ok if notifications are "lost" or if the server drops
notifications because the channel / client is too busy.

When reading a TableEntry, the client can elect to read the elapsed time
(in nanoseconds) since the entry was last hit in the data-plane. We use
a protobuf message for this field in order to enable the client to
choose whether this value should be read and returned in the
response. This is consistent with what we do for direct resource fields.

We use a repeated field of TableEntry in the notification message so
that the server can coalesce notifications in one message when
appropriate (we need to define what appropriate means in the spec).

This may not be the most comprehensive proposal, but I believe it is
sufficient for P4Runtime v1.0 as it covers the general use case.

Fixes #265

### DIFF
--- a/proto/p4/config/p4info.proto
+++ b/proto/p4/config/p4info.proto
@@ -192,7 +192,15 @@ message Table {
   // added in the future
   repeated uint32 direct_resource_ids = 7;
   int64 size = 8;  // max number of entries in table
-  bool with_entry_timeout = 9;  // entry ageing is enabled for table
+  // this enum can be extended in the future with other behaviors, such as
+  // "HARD_EVICTION"
+  enum IdleTimeoutBehavior {
+    UNSPECIFIED = 0;
+    NO_TIMEOUT = 1;
+    NOTIFY_CONTROL = 2;
+  }
+  // is idle timeout supported for this table?
+  IdleTimeoutBehavior idle_timeout_behavior = 9;
   // table with static P4 entries, cannot be modified at runtime
   bool is_const_table = 10;
 }
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -168,6 +168,18 @@ message TableEntry {
   // default action of the table. If true, the "match" field must be empty and
   // the "action" field must be populated with a valid direct action.
   bool is_default_action = 8;
+  // The TTL for the entry, in nanoseconds. A value of 0 means that the table
+  // entry never "expires".
+  int64 idle_timeout_ns = 9;
+  message IdleTimeout {
+    // Time elapsed - in nanoseconds - since the table entry was last "hit" as
+    // part of a dataplane table lookup.
+    int64 elapsed_ns = 1;
+  }
+  // Table wite: this field should be left unset.
+  // Table read: if the table supports idle timeout and time_since_last_hit is
+  // set in the request, this field will be set in the response.
+  IdleTimeout time_since_last_hit = 10;
 }
 
 // field_match_type ::= exact | ternary | lpm | range | valid
@@ -448,6 +460,7 @@ message StreamMessageResponse {
     MasterArbitrationUpdate arbitration = 1;
     PacketIn packet = 2;
     DigestList digest = 3;
+    IdleTimeoutNotification idle_timeout_notification = 4;
   }
 }
 
@@ -507,6 +520,13 @@ message Role {
   // (default case), it implies all P4 objects and control behaviors are in scope,
   // i.e. full pipeline access.
   google.protobuf.Any config = 2;
+}
+
+message IdleTimeoutNotification {
+  repeated TableEntry table_entry = 1;
+  // Timestamp at which the server generated the message (in nanoseconds since
+  // Epoch)
+  int64 timestamp = 2;
 }
 
 message Uint128 {


antoninbas:
@jafingerhut just like I did for digest, I will update the spec once we agreed on the proto interface

proto/p4/config/p4info.proto
```diff
@@ -192,7 +192,15 @@ message Table {
...
-  bool with_entry_timeout = 9;  // entry ageing is enabled for table
+  // this enum can be extended in the future with other behaviors, such as
+  // "HARD_EVICTION"
+  enum IdleTimeoutBehavior {
+    UNSPECIFIED = 0;
+    NO_TIMEOUT = 1;
+    NOTIFY_CONTROL = 2;
+  }
+  // is idle timeout supported for this table?
+  IdleTimeoutBehavior idle_timeout_behavior = 9;
```
teverman: Is a target free to fail any P4 program defining a table with NOTIFY_CONTROL at compile time if, for example, the hardware provides insufficient timeout support?

proto/p4/p4runtime.proto
```diff
@@ -168,6 +168,18 @@ message TableEntry {
   // default action of the table. If true, the "match" field must be empty and
   // the "action" field must be populated with a valid direct action.
   bool is_default_action = 8;
+  // The TTL for the entry, in nanoseconds. A value of 0 means that the table
+  // entry never "expires".
+  int64 idle_timeout_ns = 9;
```
teverman: Is this field meaningful in a table read request?

Can a target implementation reject idle_timeout_ns values that are too small for hardware support?

proto/p4/p4runtime.proto
```diff
@@ -168,6 +168,18 @@ message TableEntry {
...
+  message IdleTimeout {
+    // Time elapsed - in nanoseconds - since the table entry was last "hit" as
+    // part of a dataplane table lookup.
+    int64 elapsed_ns = 1;
+  }
+  // Table wite: this field should be left unset.
+  // Table read: if the table supports idle timeout and time_since_last_hit is
+  // set in the request, this field will be set in the response.
```
teverman: What does it mean in a streaming notification?

proto/p4/config/p4info.proto
antoninbas: I'll make that clear in the spec.  Maybe we need a general statement there about when a compiler backend should reject a program vs when to use runtime errors.

proto/p4/p4runtime.proto
antoninbas: For a read request, the server should return the value provided by the client when writing the entry.

Regarding supported `idle_timeout_ns` values, I think we should go with "best effort" for the target, instead of failing. In other words, the target should approximate the value as best as possible (kind of like for meter specs). I can add this to the spec.

proto/p4/p4runtime.proto
antoninbas: I'll update the comment for `IdleTimeoutNotifcation` to indicate that only the match key is required for members of the `table_entry` repeated field. I don't think there is value in including other fields (we probably just want to identify which entry has expired so we can remove it), including the `time_since_last_hit` field.

jafingerhut:
As far as coalescing mechanisms go, the one that you have for Digest messages might be appropriate here, too -- not the ack_timeout_ns part, but the maximum number of timeout notifications per 'batched message', and a max_timeout_ns time where that long after the first timeout notification is put into an empty buffer on the server/agent, a message to the client/controller will be generated, even if the bufer has not filled to its max capacity yet.

Do you know of existing use cases where system developers want to configure different timeouts for different table entries in the same table, or is one timeout value for all entries in the same table sufficient granularity there?  If you or anyone in the P4-API working group has such use cases, I am curious to learn a little about them if they are willing/able to share any details.

antoninbas:
@jafingerhut Thanks for the review. I think one limitation is that we do not have the ability to set table-wide properties the way P4Runtime is currently structured.

I'm not aware of any practical use case that requires different timeouts for different entries in the same table. However, I am also not aware of any limitations that would make this fundamentally hard to implement (in the worst case scenario, the switch software can pick the greatest common denominator and use this interval even though not knowing it ahead of time could be somewhat of a challenge).

I am confident that we will be able to add table-wide properties in a backward-compatible fashion should the need arise.

jafingerhut:
@antoninbas I guess it depends upon how an implementation creates these timeout notifications.  I guess no matter how it is implemented, there needs to be some kind of background task scanning through table entries to look for ones that have not been matched for the configured time or longer, since normal data plane operations cannot detect that (since the data plane accessing a table entry by definition makes it become recently accessed).

That background scanning task could physically be implemented in a variety of places, the most likely in an ASIC/FPGA/NPU implementation being the agent software, and the next most likely some kind of embedded CPU or hardware module that scans through the table entries to offload that work from the agent software.

Wherever that background scanning task is implemented, it needs to know what the timeout threshold is for all table entries, and so effectively needs to store some number of bits per installed table entry if it is per-entry-configurable.

Anyway, all of that is just me thinking out loud for why I am now coming to agree that the data plane itself shouldn't become more expensive if the timeout durations are configurable per-entry, so I drop even a shred of a resistance to that level of configurability. :-)

antoninbas:
Merging this now, I will get started on the spec today.

PR #347, opened 2018-04-24
Update code to use ID prefixes defined in p4info.proto

antoninbas:
This required major changes in the PI p4info code: because we group all
resources of the same type based on the ID prefix value, we had to
split-up direct & indirect versions of counters & meters. At the same
time, we tried to keep as much common code as possible between the
direct & indirect versions, and tried to limit the impact on the PI
public interface. As a result, the implementation does look hacky in
places (id_from_name functions in particular).


antoninbas:
Depends on https://github.com/p4lang/p4c/pull/1246

PR #349, opened 2018-04-30
Clarified usage of RECONCILE_AND_COMMIT

samar-abdi:
Will add documentation directly in the P4Runtime google doc.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -561,10 +561,16 @@ message SetForwardingPipelineConfigRequest {
     // type. Returns an error if no saved config is found or if a config is
     // provided with this message.
     COMMIT = 4;
-    // Verify, save and realize the given config. Preserve the forwarding
-    // state in the target. Returns error if no config is provided, or if the
-    // existing forwarding state in the target cannot be preserved for the
-    // given config.
+    // Verify, save and realize the given config, while preserving the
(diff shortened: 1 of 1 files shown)


proto/p4/p4runtime.proto
```diff
@@ -561,10 +561,16 @@ message SetForwardingPipelineConfigRequest {
...
-    // state in the target. Returns error if no config is provided, or if the
-    // existing forwarding state in the target cannot be preserved for the
-    // given config.
+    // Verify, save and realize the given config, while preserving the
+    // forwarding state in the target. This is an advanced use case to enable
+    // changes to the P4 forwarding pipeline configuration with minimal traffic
+    // loss. P4Runtime does not impose any constraints on the duration of the
+    // traffic loss. The support for this option is not expected to be uniform
+    // across all P4Runtime targets. A target that does not support this option
+    // may return a NOT_SUPPORTED error. For targets that support this option,
```
antoninbas: Maybe I'm not looking at the right file, but I do not see a NOT_SUPPORTED error code here: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
This file seems to indicate that UNIMPLEMENTED should be used. Could you check before we merge?

samar-abdi: Thanks for the catch. Fixed!

PR #353, opened 2018-05-15
Added support for multiple clones

samar-abdi:
This will also need change in PSA description to support multiple clones. No changes to the psa.p4 file. Note that I have changed proto field numbering. Hopefully this makes into v1.0 so we don't need to deprecate the old egress_port field.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -353,6 +353,12 @@ message PacketReplicationEngineEntry {
   }
 }
 
+// Used for replicas created for cloning and multicasting actions.
+message Replica {
+  uint64 egress_port = 1;
+  uint64 instance = 2;
+}
+
 // The (egress_port, instance) pair must be unique for each replica in a given
 // multicast group entry. A packet may be multicast by setting the
 // multicast_group field of PSA ingress output metadata to multicast_group_id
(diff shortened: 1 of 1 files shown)


jafingerhut:
@samar-abdi I have created this issue to remind us to bring this up at a future P4 Arch work group meeting: https://github.com/p4lang/p4-spec/issues/621

antoninbas:
In the referenced p4-spec issue for PSA, @jafingerhut wrote:

> Per clone session, add the option that instead of sending the clone to a single egress port, it is sent to a multicast group instead

@samar-abdi's current proposal does not include any reference to multicast, even though that would be one possible way for a target to support multiple clones. When Samar was discussing this proposal with me, I remember that we did consider this alternative, that would indeed tie the "multiple clones" mechanism to multicast:
```
 message CloneSessionEntry {
   uint64 session_id = 1;
   one_of copy {
       uint64 egress_port = 2;
       uint64 multicast_group_id = 3;
   }
   uint64 class_of_service = 4;
   uint64 packet_length_bytes = 5;
 }
```

Since the goal of this PR is to include support for multiple clones in P4Runtime v1.0 and avoid a non backward-compatible change in the near future, maybe we need to make sure everyone is on the same page and get Andy's blessing on this "implementation-agnostic" approach that doesn't explicitly refer to multicast.

samar-abdi:
@antoninbas @jafingerhut 
My only concern with a reference to a multicast group is the extra care it needs when adding/removing the clone session entry. Nothing wrong with it, just that in our experience such references (like flows/groups/members) are a source of bugs because either the controller or the switch missed corner cases in the management of the referenced object.

antoninbas:
I agree with @samar-abdi 
Since there is no way to predict the future PSA discussions and if we want this to be in P4Runtime v1.0, I see no reason not to proceed with merging this.

jafingerhut:
@samar-abdi @antoninbas Sorry for not replying earlier.  I would prefer not to unilaterally make changes to the PSA spec, and I don't have a good guess how the conversation will go on this topic.

I suspect if the P4Runtime API uses something like 'repeated' on '(egress_port, instance)' value pairs, or just 'egress_port' values alone, then if a PSA implementation was able to implement this by using one multicast group for each such clone session "under the hood", that should not be difficult for them to manage in their agent software.

samar-abdi:
This PR should not have any impact on current PSA. As mentioned earlier, the P4Runtime spec will mention that only one replica can be created for a clone session in the current PSA version. If PSA evolves to support multiple clone replicas, we can have a subsequent P4Runtime minor release, which may remove this restriction or add a different restriction on the number of clones.

Since we want to quickly close on both P4Runtime 1.0 proto and spec, can we move forward with merging this PR?

antoninbas:
I'm merging this now. As long as PSA does not _mandate_ that creating multiple clones must be done through multicast, we should not need to change this down the line. I expect some targets will achieve multiple clones by creating a multicast group per session, which - as @jafingerhut mentioned - should be possible to do in the agent.

hanw:
@antoninbas Does P4Runtime support specifying both egress port and multicast group in the same clone session?

antoninbas:
P4Runtime takes as parameter a set of unicast ports for a clone session. This **may** be implemented by the switch stack through the use of a multicast group. The API for defining cloning sessions doesn't mention multicast groups.
See https://github.com/p4lang/p4runtime/blob/master/proto/p4/v1/p4runtime.proto#L420

PR #356, opened 2018-05-19
Add major version number to P4Runtime protobuf packages

### LINKED ISSUE #294: P4Runtime API versioning
@aghaffarkhah @wmohsin this is something we discussed offline but I cannot find any webpage describing this. Could you give an example of a protobuf file with version info?

### LINKED ISSUE #350: proto file definition is causing import loop in golang
I recently updated the .proto files and regenerated my golang grpc code. I'm hitting this issue:

p4/p4runtime.proto and p4/p4types.proto form "package p4" in golang
p4/config/p4info.proto  forms "package p4.config" in golang

package p4 is importing p4/config
p4/config is importing p4 

This is creating an import loop when I compile my go binaries.

antoninbas:
We follow best practices for protobuf cloud services and add the major
version number to the package path. This change affects many C++ source
files in this project and I didn't find a good way to make the diff
smaller by using namespace "tricks".

Fixes #294

This commit also split p4/p4types.proto into p4/v1/p4data.proto and
p4/config/v1/p4types.proto, thus breaking the circular import in golang
between the p4 and p4.config packages.

Fixes #350

### DIFF
--- a/proto/demo_grpc/simple_router_mgr.cpp
+++ b/proto/demo_grpc/simple_router_mgr.cpp
@@ -35,6 +35,9 @@
 
 #include "google/rpc/code.pb.h"
 
+namespace p4rt = ::p4::v1;
+namespace p4config = ::p4::config::v1;
+
 using grpc::ClientContext;
 using grpc::Status;
 using grpc::ClientReaderWriter;
@@ -140,12 +143,12 @@ struct CounterQueryHandler : public MgrHandler {
   CounterQueryHandler(SimpleRouterMgr *mgr,
                       const std::string &counter_name,
                       size_t index,
-                      std::promise<p4::CounterData> &promise)
+                      std::promise<p4rt::CounterData> &promise)
       : MgrHandler(mgr), counter_name(counter_name), index(index),
         promise(promise) { }
 
   void operator()() {
-    p4::CounterData d;
+    p4rt::CounterData d;
     int rc = simple_router_mgr->query_counter_(counter_name, index, &d);
     if (rc) {
       d.set_packet_count(
@@ -156,7 +159,7 @@ struct CounterQueryHandler : public MgrHandler {
 
   std::string counter_name;
   size_t index;
-  std::promise<p4::CounterData> &promise;
+  std::promise<p4rt::CounterData> &promise;
 };
 
 struct ConfigUpdateHandler : public MgrHandler {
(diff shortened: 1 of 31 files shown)


proto/demo_grpc/simple_router_mgr.cpp
```diff
@@ -35,6 +35,9 @@
 
 #include "google/rpc/code.pb.h"
 
+namespace p4rt = ::p4::v1;
```
wmohsin: I feel p4v1 is a better fit. First, its the same number of letters :) But more importantly, the current p4rt is de-emphasizing that these are v1 messages. In the future, when we have v2, we would need to introduce a new alias, and p4v2 would be better than something like p4rt2. Also, at that point, this p4rt would become ambiguous wrt which version it refers to. Finally, re-pointing p4rt to ::p4::v2 will be a no-no (IMHO) since they will be incompatible by definition.
I would say the same for config (p4configv1).
What do you think?

antoninbas: If you are referring specifically to the demo client under demo_grpc, then I agree it should probably use `p4::v1` or `p4v1` instead of `p4rt`.
However if this is a generic remark that includes the server implementation files, I am not sure I agree. Actually the reason why I introduced the `p4rt` alias was to make potential future changes (including v1 -> v2) easier to implement. Even though this is far in the future, I would expect most messages to stay backward-compatible with the change from v1 to v2, and the namespace alias was meant to help keep the CL small.
That being said, there is no urgency to merge this (you also seem to agree with the .proto changes, which is the probably the most important thing to get right at the moment), so I can take some more time to think about this.

wmohsin: I thought about this a little more, and looked at code which follows the guidelines where are trying to adhere to. I do style similar to yours (p4rt=::p4::v1) as well as variants (p4rtv1, p4rtv1, p4rt_v1, etc.). I guess there is no hard and fast rule around this. LGTM.

antoninbas:
@wmohsin I updated the namespace aliases. Even though the fact that the namespace aliases are only used within implementation files means that there should be no name collisions, I can't think of a very good reason not to include the version in the namespace alias. Since we are not able to predict the future, it seems like the safest choice.

antoninbas:
While we could place some stable message definitions under p4/types (as suggested by https://cloud.google.com/apis/design/directory_structure), such as `Uint128`, `Index` and possibly `Error`, it is probably better to get some mileage out of P4Runtime first. I'm not sure it would also be a huge help to implementers, given that the number of messages that would be safe to place under p4/types is probably very small.

If you think p4/types is a good idea, please comment here.

Issue #359, opened 2018-05-25
Use uint32 instead of uint64 for egress_port in P4Runtime multicast Replica message


antoninbas:

samar-abdi:
We can also use uint32 for replica ID, while we are at it. 64 bits seems excessive.

antoninbas:
This has been fixed

PR #364, opened 2018-05-30
Added reserved ports and fixed metadata types to uint32

samar-abdi:
uint32 types for PSA metadata are compatible with the max suggested bitwidths for clone session, multicast group, CoS and replica ID.

### DIFF
--- a/proto/p4/p4runtime.proto
+++ b/proto/p4/p4runtime.proto
@@ -344,6 +344,20 @@ message CounterData {
 }
 
 //------------------------------------------------------------------------------
+// Reserved controller-specified port numbers for reference.
+// Wrapping the enum in a message to avoid name collisions in C++, where "enum
+// values are siblings of their type, not children of it"
+message Ports {
+  enum ReservedPorts {
+    // Port number 0 in invalid. Egress port is initialized to this
+    // controller port number at the beginning of ingress and should
+    // be assigned a valid number in the ingress for all unicast packets.
+    P4RUNTIME_PORT_UNKNOWN = 0;
+    P4RUNTIME_PORT_CPU = 0xFFFFFFFD;
+    P4RUNTIME_PORT_RECIRCULATE = 0xFFFFFFFE;
+  }
+}
+
 // Only one instance of a Packet Replication Engine (PRE) is expected in the
 // P4 pipeline. Hence, no instance id is needed to access the PRE.
 message PacketReplicationEngineEntry {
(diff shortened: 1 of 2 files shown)


proto/p4/p4runtime.proto
```diff
@@ -344,6 +344,20 @@ message CounterData {
 }
 
 //------------------------------------------------------------------------------
+// Reserved controller-specified port numbers for reference.
+// Wrapping the enum in a message to avoid name collisions in C++, where "enum
+// values are siblings of their type, not children of it"
+message Ports {
+  enum ReservedPorts {
+    // Port number 0 in invalid. Egress port is initialized to this
```
antoninbas: s/in/is

proto/p4/p4runtime.proto
```diff
@@ -384,10 +398,10 @@ message MulticastGroupEntry {
 // packet_length_bytes field is 0, no truncation on the clone(s) will be
 // performed.
 message CloneSessionEntry {
-  uint64 session_id = 1;
+  uint32 session_id = 1;
   repeated Replica replicas = 2;
-  uint64 class_of_service = 3;
-  uint64 packet_length_bytes = 4;
+  uint32 class_of_service = 3;
+  uint32 packet_length_bytes = 4;
```
antoninbas: Isn't the best practice to use int32 here? I'm not totally sure but I think I remember @wmohsin mentioning something like this.

proto/p4/p4runtime.proto
```diff
@@ -344,6 +344,20 @@ message CounterData {
...
 
 //------------------------------------------------------------------------------
+// Reserved controller-specified port numbers for reference.
+// Wrapping the enum in a message to avoid name collisions in C++, where "enum
+// values are siblings of their type, not children of it"
+message Ports {
+  enum ReservedPorts {
+    // Port number 0 in invalid. Egress port is initialized to this
+    // controller port number at the beginning of ingress and should
+    // be assigned a valid number in the ingress for all unicast packets.
```
jafingerhut: PSA v1.0 defines that drop=true for all packets when ingress processing begins, and if you want the packet to be sent unicast to a particular egress port, you must change both drop=false and assign a value to egress_port, which in PSA has no initialized value (i.e. a portable PSA program cannot assume it will be initialized to 0, or any other particular value).  See Section 4.2 of the PSA spec and comments on the metadata fields there.  You are welcome to repeat those details here, but simply leaving this comment as '// Port number 0 is invalid.' might be enough here.

antoninbas:
Fixes #359

antoninbas:
@samar-abdi what's the status of this?
I think @wmohsin didn't like the enum wrapping and made changes in the spec.

samar-abdi:
Synced now with the spec. PTAL.

antoninbas:
@samar-abdi the proto file is rejected by the protoc compiler: it looks like it doesn't like such large unsigned integers for enum values

antoninbas:
Superseded by https://github.com/p4lang/p4runtime/pull/52

PR #368, opened 2018-06-01
Remove VALID match type from P4Runtime

### LINKED ISSUE #367: Remove valid match type from p4info & p4runtime
As decided at the 5/30/2018 P4 API WG meeting.

antoninbas:
We remove the VALID match type from p4runtime.proto and
p4info.proto. The valid match type does not exist in P4_16, which means
that the p4c compiler never produces a VALID match type in P4Info even
for P4_14 programs, since all P4_14 programs are converted to P4_16
before P4Info is generated.

While the change is not backward-compatible since we are deleting
Protobuf fields, it should not impact existing Protobuf binaries that do
not use VALID, as it does not affect other Protobuf fields (we did not
change any of the Protobuf integer field values).

The VALID match type is preserved in the PI implementation for now, to
ensure backward-compatibility with the bmv2 JSON produced by the legacy
compiler (p4c-bm).

Fixes #367

### DIFF
--- a/proto/frontend/src/device_mgr.cpp
+++ b/proto/frontend/src/device_mgr.cpp
@@ -639,10 +639,11 @@ class DeviceMgrImp {
       switch (finfo->match_type) {
         case PI_P4INFO_MATCH_TYPE_VALID:
           {
-            auto valid = mf->mutable_valid();
+            auto exact = mf->mutable_exact();
             bool value;
             mk_reader.get_valid(finfo->mf_id, &value);
-            valid->set_value(value);
+            exact->set_value(
+                value ? std::string("\x01", 1) : std::string("\x00", 1));
           }
           break;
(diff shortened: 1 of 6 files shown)


antoninbas:
@teverman this should match what we agreed on by email

proto/frontend/src/device_mgr.cpp
```diff
@@ -639,10 +639,11 @@ class DeviceMgrImp {
       switch (finfo->match_type) {
         case PI_P4INFO_MATCH_TYPE_VALID:
           {
-            auto valid = mf->mutable_valid();
+            auto exact = mf->mutable_exact();
```
teverman: Can you add some comments?  I assume this is for backward compatibility, bit it would help to clarify.

antoninbas: @teverman Done. PTAL.

PR #369, opened 2018-06-01
Add static max group size for action profile to P4Info

### LINKED ISSUE #366: Add static action profile maximum group size to P4Info

antoninbas:
This information must be known at compile-time. The dynamic group max
size provided at hroup creation time is now optional and must never
exceed the static max group size when provided.

Fixes #366


antoninbas:
@aghaffarkhah could you take a look and let us know if it works for you, this should match the discussion we had by email

PR #370, opened 2018-06-04
Update p4runtime.proto

wmohsin:
Adds atomicity semantics (continue-on-error, rollback-on-error, dataplane-atomic) for a Write batch.

### DIFF
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -62,7 +62,7 @@ message WriteRequest {
   uint64 role_id = 2;
   Uint128 election_id = 3;
   // The write batch, comprising a list of Update operations. The P4Runtime
-  // server may arbitrarily reorder message within a batch to maximize
+  // server may arbitrarily reorder messages within a batch to maximize
   // performance.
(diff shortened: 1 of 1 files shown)


proto/p4/v1/p4runtime.proto
```diff
@@ -61,7 +61,35 @@ message WriteRequest {
...
+    // back such that both software and dataplane state is consistent with the
+    // state before the batch was attempted. The resulting behavior is
+    // all-or-none, except the batch is not atomic from a data plane point of
+    // view. Every dataplane packet is guaranteed to be processed according to
+    // table contents as they are between two individual operations of the
+    // batch, but there could be several packets processed that see each of
+    // these intermediate stages.
+    ROLLBACK_ON_ERROR = 1;
+    // Optional. Every dataplane packet is guaranteed to be processed according
+    // to table contents before the batch began, or after the batch completes.
```
aghaffarkhah: s/before the batch began, or after the batch completes/right before or after the batch was programmed to the HW

wmohsin: Done

PR #374, opened 2018-06-07
Support for arch-specific extensions for tables in P4Runtime

### LINKED ISSUE #301: Make sure that P4 Runtime can be easily extended for non-PSA architectures
A few TODO items that I'm aware of:
1. Read / Write support for arbitrary architecture-specific externs. There is already some basic support but that hasn't been tested with a concrete architecture.
2. Support additional match types. This was proposed by @jtardo here (https://github.com/p4lang/PI/issues/241#issuecomment-347966628).
3. Ability to extend `TableEntry` message with a `proto.Any` message field. That will provide support for architecture-specific runtime table properties and provide support for architecture-specific direct resources.
4. Support for different types of notifications on bi-directional stream.

As mentioned above we already have some tentative - and undocumented - support for 1. in p4runtime.proto and p4info.proto. I also have a few ideas for 2. and 3. For now I am just opening this issue to make sure it makes it to the roadmap for v1.0.

antoninbas:
This introduces support for:
 - additional architecture-specific match types. This is achieved on the
 P4Info side by wrapping the match_type field in a oneof, where the
 alternative is a string specifying the arch-specific match-type
 name. On the P4Runtime side, we simply add Any as a possible value for
 the FieldMatch oneof.
 - ability to extend TableEntry message with Any message field.

Fixes #301

### DIFF
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -240,6 +240,9 @@ message FieldMatch {
     Ternary ternary = 3;
     LPM lpm = 4;
     Range range = 6;
+    // Architecture-specific match value; it corresponds to the other_match_type
+    // in the P4Info MatchField message.
+    google.protobuf.Any other = 7;
   }
 }
 
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -242,7 +242,7 @@ message FieldMatch {
     Range range = 6;
     // Architecture-specific match value; it corresponds to the other_match_type
     // in the P4Info MatchField message.
-    google.protobuf.Any other = 7;
+    google.protobuf.Any other = 100;
   }
 }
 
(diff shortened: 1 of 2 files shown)


antoninbas:
I don't know how controversial this is going to be. AFAIK wrapping the `match_type` enum in a oneof is backward-compatible, but I could be wrong (see "Backward-compatibility issues" in https://developers.google.com/protocol-buffers/docs/proto3#oneof).

proto/p4/v1/p4runtime.proto
```diff
@@ -240,6 +240,9 @@ message FieldMatch {
     Ternary ternary = 3;
     LPM lpm = 4;
     Range range = 6;
+    // Architecture-specific match value; it corresponds to the other_match_type
+    // in the P4Info MatchField message.
+    google.protobuf.Any other = 7;
```
teverman: Might be a good idea to make this field ID 8 since valid used to be 7.

samar-abdi: Or 100 to give space for other intrinsic types.

antoninbas: done

antoninbas:
Thanks for trying this out @teverman and thanks for the reviews

PR #380, opened 2018-06-21
Support arbitrary varbits in P4Runtime

### LINKED ISSUE #371: Support varbit<W> properly in P4Data
We should include the dynamic-length (or restrict support to multiples of 8 bits?). It's hard to think of a use-case that would require exposing varbits through P4Runtime (maybe digests?), but there is very little cost to add this support.

antoninbas:
We include the dynamic size in the P4Data message for varbit
types. While we can only think of very few cases where varbits would
need to be exposed to the control-plane - and even fewer are realistic -
this seems to be the most future-proof approach.

Fixes #371

### DIFF
--- a/proto/p4/v1/p4data.proto
+++ b/proto/p4/v1/p4data.proto
@@ -18,18 +18,24 @@ package p4.v1;
 
 message P4Data {
   oneof data {
-    bytes bitstring = 1;  // for bit<W>, int<W>, varbit<W>
-    bool bool = 2;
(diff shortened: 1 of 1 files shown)


antoninbas:
I took the liberty of re-numbering the proto fields (mostly for aesthetics reasons) since I'm under the impression that no-one is really using P4Data yet. If someone has an objection, I don't mind reverting that part of the commit.

PR #383, opened 2018-06-25
Use highest election_id in MasterArbitrationUpdate response

ccascone:
Spec says:

> After the controller sends a StreamMessageRequest message to the P4Runtime server, the server sends a StreamMessageResponse message back to the controller, in which it populates the MasterArbitrationUpdate message using the device_id, role, and election_id it previously received from all the controllers via the StreamMessageRequest message. **The election_id populated in the response is always the highest election_id, referring to the current master after the request was processed.**

### DIFF
--- a/proto/server/pi_server.cpp
+++ b/proto/server/pi_server.cpp
@@ -288,7 +288,8 @@ class DeviceState {
       to->set_high(from.high());
       to->set_low(from.low());
     };
-    convert_u128(connection->election_id(), arbitration->mutable_election_id());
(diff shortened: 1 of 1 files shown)


ccascone:
How do I run gtests? Did `make check` and tests there were passing.

antoninbas:
make check runs them, Travis will fail if one test is broken anyway

antoninbas:
@ccascone could you fix the style issue so we can merge this?

PR #384, opened 2018-06-26
Add support for serializable enums in P4Runtime

antoninbas:
Proposal for supporting P4 serializable enums in P4Runtime. For each
serializable enum, P4Info includes the underlying value for each named
member of the enum. On the P4Runtime side, P4Data enables providing
either the name or the numerical value (it should be possible to provide
a value for a serializable enum even if this value does not correspond
to named member, as long as the numerical value falls into the range of
the underlying representation).

See P4_16 spec section 7.2.1 (Enumeration types) for more background on
safe enums vs. serializable enums.

### DIFF
--- a/proto/p4/config/v1/p4types.proto
+++ b/proto/p4/config/v1/p4types.proto
@@ -32,13 +32,15 @@ package p4.config.v1;
  * | error              | error     | error        | allowed         |
  * | match_kind         | error     | error        | error           |
  * | bool               | error     | error        | allowed         |
- * | enum               | error     | error        | allowed         |
+ * | enum               | allowed*  | error        | allowed         |
  * | header             | error     | allowed      | allowed         |
  * | header stack       | error     | error        | allowed         |
  * | header_union       | error     | error        | allowed         |
  * | struct             | error     | error        | allowed         |
  * | tuple              | error     | error        | allowed         |
  * |--------------------|-----------|--------------|-----------------|
+ *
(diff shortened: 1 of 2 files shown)


antoninbas:
@chrispsommers @jafingerhut this may interest you as well

proto/p4/config/v1/p4types.proto
```diff
@@ -162,6 +163,20 @@ message P4EnumTypeSpec {
...
 }
 
+// For serializable (or "unsafe") enums, which have an underlying type. Note
+// that as per the P4_16 specification, the underlying representation can only
+// be a bit<W> type.
+message P4SerializableEnumTypeSpec {
+  message Member {
+    string name = 1;
+    bytes value = 2;
+    repeated string annotations = 3;
```
chrispsommers: This might apply to other areas of this file as well: in addition to generic `annotations`, should we consider adding an optional `Documentation` field for type definitions? For example, for wither safe or serializable enums I can imagine adding brief and sometimes detailed descriptions of the meaning and usage of each value. Initially they could come from @brief and @description annotations; eventually they could come from javadoc style comments. Again, this is all in the spirit of making the P4Info contain a lot of the needed control plane details.

chrispsommers: Fair enough!

antoninbas:
merging this since there doesn't seem to be any controversy

PR #385, opened 2018-06-27
Reject priority 0 for ternary matches in P4Runtime

antoninbas:
As per the P4 spec


antoninbas:
As discussed and agreed upon at the 06/20/2018 P4 API WG meeting

PR #387, opened 2018-06-27
Add device manager general error message

loalan:
Depending on the data the controller sends over, the response code sometimes comes back without a string message, just an UNKNOWN code. For a stack of these errors, set the last error message to the error status. This will give the controller some idea what failed.


loalan:
Ok, I'll take a deeper look and a different approach when I get a chance. The problem is at the server side thought, when getting this single Status object, there message for the "unknown" error code, which makes it pretty difficult to understand if this is the only error.

antoninbas:
The actual error codes for each operation in the batch are actually in the `details` field. I think this is described in the P4Runtime spec.

PR #388, opened 2018-06-27
Expose the p4info to proto conversion library

loalan:
When writing a test controller built from a different repo (for a vendor specific PI target), libpiconvertproto library is very useful, but not exposed outside of p4lang/PI. It would be nice to have this available with the other PI libraries from a make install. (I didn't see an easier way of doing this without adding it to the install)


proto/p4info/Makefile.am
```diff
@@ -9,11 +9,18 @@ bin_PROGRAMS = pi_convert_p4info
...
 # protobuf message
-noinst_LTLIBRARIES = libpiconvertproto.la
+lib_LTLIBRARIES = libpiconvertproto.la
 
 libpiconvertproto_la_SOURCES = \
 p4info_to_and_from_proto.cpp \
-p4info_to_and_from_proto.h
+p4info_to_and_from_proto.h 
+
+#nobase_include_HEADERS = PI/proto/p4info/p4info_to_and_from_proto.h
```
antoninbas: If the header is not installed, how are you going to use the library?

proto/p4info/Makefile.am
```diff
@@ -9,11 +9,18 @@ bin_PROGRAMS = pi_convert_p4info
...
+
+libpiconvertproto_la_LIBADD = \
```
antoninbas: I may be missing something, but I think the dependency is only on protobuf, not grpc

proto/p4info/Makefile.am
```diff
@@ -23,3 +30,4 @@ $(builddir)/libpiconvertproto.la \
 $(top_builddir)/libpiprotobuf.la \
 $(top_builddir)/../src/libpip4info.la \
 $(PROTOBUF_LIBS)
+
```
antoninbas: new empty line?

loalan:
Removed references to grpc in the makefile. Installed the header under PI/proto, hope that makes sense.

loalan:
Now with changes to PI test and other components that use the conversion internally

proto/demo_grpc/Makefile.am
```diff
@@ -8,7 +8,7 @@ AM_CPPFLAGS = \
 -I$(top_builddir)/grpc_out \
 -I$(top_srcdir)/p4info
 
-AM_CXXFLAGS = -Wall -Werror
+AM_CXXFLAGS = -Wall -Werror -Wno-unused-command-line-argument
```
antoninbas: Could you add a comment explaining the `-Wno-unused-command-line-argument` addition?

proto/p4info/Makefile.am
```diff
@@ -9,11 +9,17 @@ bin_PROGRAMS = pi_convert_p4info
 
 # convenience library to convert back-and-forth between p4info and the P4Info
 # protobuf message
-noinst_LTLIBRARIES = libpiconvertproto.la
+lib_LTLIBRARIES = libpiconvertproto.la
 
 libpiconvertproto_la_SOURCES = \
 p4info_to_and_from_proto.cpp \
-p4info_to_and_from_proto.h
+p4info_to_and_from_proto.h
```
antoninbas: this line can be removed

proto/demo_grpc/Makefile.am
loalan: linking failed since -lpthread was an unused linker argument. 
clang: error: argument unused during compilation: '-pthread' [-Werror,-Wunused-command-line-argument]
But I looked at this more, seems to be just a problem with xcode/clang. It's just on the demo grpc code, let me know what's preferred

proto/p4info/Makefile.am
loalan: ok, added a ws at the end, I'll fix it.

antoninbas: the line is actually not needed because the file is now listed in `_HEADERS` so it is an implicit source

proto/demo_grpc/Makefile.am
antoninbas: I think you should try removing the `AM_CXXFLAGS` line altogether. The configure.ac should make sure the flags are correct already.

proto/p4info/Makefile.am
loalan: ok

proto/demo_grpc/Makefile.am
loalan: Just tried that, now getting multiple warnings->error in the pb header:
../grpc_out/p4/v1/p4runtime.grpc.pb.h:210:59: error: unused parameter 'service' [-Werror,-Wunused-parameter]

loalan: removing AM_CXXFLAGS shows that -Wextra is now on the compile line, which causes the detection of the unused parameter errors

antoninbas:
@loalan ok I'll merge the PR as is and I'll look into fixing the warnings later

PR #406, opened 2018-09-20
Support for one-shot action selector programming mode

antoninbas:
P4Runtime has introduced a new mode to enable easier programming of
indirect tables which use an action selector. This new mode is called
"one shot". This commit introduces support for this in the reference
P4Runtime implementation built on top of PI.

Because one P4Runtime operation for one-shot programming may translate
to many PI function calls, we add the ability to do some rollback in
case of failure. This is not as powerful as the P4Runtime transation
semantics, but it should work for simple operations, like rolling back
creation of members & groups.

We chose to implement one-shot table MODIFY by first creating a new
group and adding the necessary members, then swapping the old group
pointer with the new one in the table entry. This means that we need
enough space in memory to hold the extra group. Note that the current
implementation always creates a group, even if the P4Runtime
ActionProfileActionSet only includes a single action.

The specification states that the 2 programming modes cannot be used
simultaneously for a given action selector. This commit implements this
check. If the action selector becomes empty, we enable the client to
switch modes.

Just like for "manual" action selector programming, we do not have any
support for weight and watch yet.

Finally this commit also adds the ASSERT_OK and EXPECT_OK macros for
gTest which enable us top check if the Status object returned by
DeviceMgr is OK in a less verbose fashion. Note that existing unit tests
haven't been updated yet to use these new macros.

### DIFF
--- a/proto/frontend/src/action_prof_mgr.cpp
+++ b/proto/frontend/src/action_prof_mgr.cpp
@@ -71,6 +71,11 @@ ActionProfBiMap::remove(const Id &id) {
   }
 }
 
+bool
+ActionProfBiMap::empty() const {
+  return bimap.empty();
+}
+
 ActionProfGroupMembership::ActionProfGroupMembership() { }
 
 void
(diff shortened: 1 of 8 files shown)


jafingerhut:
Is it allowed to be able to do multiple 'one shot' updates in a single P4Runtime batch operation?  I ask primarily because wondering if that might make it tricky to implement the optional ROLLBACK_ON_ERROR batch mode.  I have not yet verified that there is definitely something tricky there -- wondering aloud right now.

antoninbas:
@jafingerhut I don't see why it shouldn't be possible. However I have no plans to implement `ROLLBACK_ON_ERROR` in this repository for arbitrary targets. As you pointed out some time ago during the WG discussions, it is actually not possible to do `ROLLBACK_ON_ERROR` at this layer for exact match tables.
The limited rollback support at this layer is meant for P4Runtime operations that map to many PI function calls and should be reserved for PI calls that can be reverted unconditionally (for now it is only used for reverting *insert* operations for *action profiles* - I would expect that `insert` operations on their own can always be reverted...). An implementation for a specific target whose SDK supports arbitrary `ROLLBACK_ON_ERROR` operations could in theory leverage that instead of the rollback code I'm checking in now.

antoninbas:
@konne-google it should be in device_mgr.cpp
Github collapses the file in the diff view if the size of the diff for the file is too high so your search may have missed it

konne-google:
Yes, github collapsing the file was the problem. Thanks Antonin!

PR #413, opened 2018-10-15
Add needed package libboost-thread-dev

jafingerhut:


jafingerhut:
When I tried following build instructions for simple_switch_grpc, which said to follow instructions to install p4lang/PI, the install steps for installing PI failed at the configure command, unless I first installed libboost-thread-dev.  I am guessing here based on the configure script failing with these messages, when I did not have libboost-thread-dev package installed:
```
checking for boostlib >= 1.54.0… yes
checking whether the Boost::Thread library is available.. yes
configure: error: Could not find a version of the library!
configure: error: ./configure failed for photo
```

jafingerhut:
Done.

PR #414, opened 2018-10-18
Fix typos for comments in include/PI/pi_learn.h

Yukinarii:


antoninbas:
Hi @Yukinarii 
Thanks for your patch. Could you confirm that you or your organizations are p4.org members? It is a requirement for me to accept this PR.

Yukinarii:
Hi @antoninbas 
Actually I am a student from NCTU which is a p4.org university member, but I don't know if there is any approach to prove this.

PR #422, opened 2018-12-02
Update supported P4Runtime version to v1.0.0-rc4

### LINKED ISSUE #416: simple_switch_grpc allows setting a table's default action to an action with @tableonly attribute
If this issue belong in a different repo (e.g. p4lang/behavioral-model), please let me know and I will close this one and create a new one there.

This was done using Python bindings of the P4Runtime API over gRPC.

I have tested this with about a 1-week-old version of p4lang/p4c, behavioral-model, p4runtime, and PI repositories, so if things have changed since then, let me know and I will update.

From my reading of the P4_16 spec, actions with the @tableonly attribute should be rejected if the control plane attempts to set such an action as a table's default action.

I have also seen simple_switch_grpc allow adding a table entry with an action that has the @defaultonly attribute, which similarly seems like it ought to cause an error.

antoninbas:
We also add support in PI for the different table action scopes
(TABLE_AND_DEFAULT, TABLE_ONLY and DEFAULT_ONLY). Proper checks are
performed in the P4Runtime DeviceMgr to ensure that the client doesn't
violate the scope.

Fixes #416


include/PI/p4info/tables.h
```diff
@@ -104,12 +115,19 @@ const pi_p4_id_t *pi_p4info_table_get_actions(const pi_p4info_t *p4info,
...
 
+//! Returns NULL if @action_id is not valid for the table.
+const pi_p4info_action_info_t *pi_p4info_table_get_action_info(
+    const pi_p4info_t *p4info, pi_p4_id_t table_id, pi_p4_id_t action_id);
+
 bool pi_p4info_table_has_const_default_action(const pi_p4info_t *p4info,
                                               pi_p4_id_t table_id);
 
+// has_mutable_action_params is deprecated
+// it will always be set to false if the table has a const default action, true
```
jafingerhut: Question: Did you intend to day "true" instead of "false" in this line?

If not, can you explain why?

antoninbas: No worries. I actually got it wrong the first time I wrote the comment and realized it only when submitting the PR :)

PR #437, opened 2019-01-20
Fix gRPC failure and demo_grc

hesingh:
Fix Issue # 436


jafingerhut:
Closing this PR, as the issue mentioned was closed as fixed some time ago, in a different way, and the install script included in this PR is many years old and probably no longer works.

PR #448, opened 2019-02-15
Update P4Runtime to v1.0.0

antoninbas:
* Update P4Runtime to v1.0.0

* Better handling of action profile member weights


antoninbas:
@ccascone if you want to make sure this won't break ONOS

ccascone:
It shouldn't, as ONOS already set group member weights to 1

PR #450, opened 2019-03-07
Fix Bazel build

antoninbas:
Added missing PI C source files.
We now use a glob (with some excluded files) for the PI src files, as in
the absence of CI for the Bazel build, we keep forgetting to update the
srcs. Of course the right solution would be to add CI...


antoninbas:
Merging right away since independent of Travis CI

PR #456, opened 2019-03-27
Add selector groups size checks as per P4Runtime specification

### LINKED ISSUE #452: Implement proper rules to set group max_size as per the spec
https://s3-us-west-2.amazonaws.com/p4runtime/docs/v1.0.0/P4Runtime-Spec.html#sec-max-size-rules

currently if max_size is set to 0 in the `ActionProfileGroup` message, we just pass it through to PI as 0. We completely ignore max_group_size from P4Info...

antoninbas:
We enforce everything we can in the DeviceMgr implementation and return
the appropriate error code. If both the dynamic max size (provided at
group creation time) and the statis max size (from P4Info) are 0,
pi_act_prof_grp_create is called with a max_size of 0. It is the
responsibility of the PI target implementation to behave accordingly to
the spec then ("the target should use the maximum value it can
support").

Fixes #452

### DIFF
--- a/include/PI/p4info/act_profs.h
+++ b/include/PI/p4info/act_profs.h
@@ -54,6 +54,9 @@ bool pi_p4info_act_prof_is_action_of(const pi_p4info_t *p4info,
 size_t pi_p4info_act_prof_max_size(const pi_p4info_t *p4info,
                                    pi_p4_id_t act_prof_id);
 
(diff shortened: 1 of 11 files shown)


antoninbas:
@ccascone you may want to check that nothing here comes as a surprise

PR #462, opened 2019-04-10
Deprecate temporary P4DeviceConfig message

antoninbas:
The assumption was always that this was a temporary message. It was
required when using bmv2 simple_switch, but is no longer needed with
simple_switch_grpc.

As a result of this change, the
ForwardingPipelineConfig::p4_device_config binary string is passed on
directly to PI. For backward compatibility, the DeviceMgr still tests
whether this field is a serialized P4DeviceConfig message (and behaves
as before).

### DIFF
--- a/proto/demo_grpc/pi_server_main.cpp
+++ b/proto/demo_grpc/pi_server_main.cpp
@@ -19,6 +19,7 @@
  */
 
 #include <PI/frontends/proto/device_mgr.h>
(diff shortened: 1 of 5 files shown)


stolsma:
Yeah, in favour of this change... Took us a whole day to find out that this P4DeviceConfig construct was needed when prototyping p4runtime solutions😢

PR #465, opened 2019-04-23
Restrict Python interpreter detection to python2.7

### LINKED ISSUE #464: configure step on latest Ubuntu 18.04 detect Python3, not Python2
This is not an urgent issue, but wanted to file one with some details in case it is a sign of some changes that ought to be made some time.

I have this bash script that I have run about once per month on an Ubuntu 16.04 and 18.04 Linux system, freshly installed, but always with most recent updates from Ubuntu before I run it each time.

When I ran it on 2019-Apr-21, for the first time I have run it I saw a weird error message:
```
 /usr/bin/install -c -m 644 gen_pd.py tenjin_wrapper.py '/usr/local/lib/python3.6/site-packages'
Byte-compiling python modules...
gen_pd.pytenjin_wrapper.py
  File "/usr/local/lib/python3.6/site-packages/gen_pd.py", line 99
    print "cannot have 2 different lpm matches in a single table"
                                                                ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print("cannot have 2 different lpm matches in a single table")?

Byte-compiling python modules (optimized versions) ...
gen_pd.pytenjin_wrapper.py
```
Earlier I noticed that the p4lang/PI configure script run found python3, not python (version 2), which did not happen in 2019-Mar-14 when I ran it last.  I do not know whether this change is because of changes in Ubuntu 18.04, in p4lang/PI, or some combination of the two.

Earlier during the run of the configure script I saw these lines of output:
```
checking for a Python interpreter with version >= 2.7... python3
checking for python3... /usr/bin/python3
checking for python3 version... 3.6
```
During the run on 2019-Mar-14, the configure script found /usr/bin/python (Python 2), and there was thus no error running the Python program.

I will add links to the full install log output soon, in case it is useful.

antoninbas:
On recent systems, python defaults to python3 which triggers an error
message during the installation phase. A poor man's solution is to force
configure to pick python2.

Fixes #464

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -17,8 +17,12 @@ AC_PROG_CC_C99
 # long as --with-bmv2 / --with-fe-cpp / --with-proto is not specified of course)
 AC_PROG_CXX
 
-# TODO: make this conditional on the PD if possible
-AM_PATH_PYTHON([2.7],, [:])
+# TODO: this can be removed when the Python scripts are updated to work with
+# python3 or when the PD generator code is removed altogether.
+m4_define_default([_AM_PYTHON_INTERPRETER_LIST], [python2 python2.7])
+AM_PATH_PYTHON([2.7],, [
+    AC_MSG_WARN([Cannot detect python2.7, is it installed? As a result we will skip installation of the PD generator (deprecated).])])
+AM_CONDITIONAL([HAVE_PYTHON2], [test "$PYTHON" != :])
 
 DX_DOT_FEATURE([ON])
 DX_PDF_FEATURE([ON])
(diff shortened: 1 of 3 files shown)


jafingerhut:
Does this also imply that a prerequisite for installing on Ubuntu 18.04 and similar systems should be to install a Python2 package?  Python2 is not installed by default on a new Ubuntu 18.04 system.

antoninbas:
I didn't think of that.
But yes python2 is basically still a dependency for all the p4lang software in general. I'm going to amend my PR to print a warning if python2 is not detected and skip the PD generator (that code is unused AFAIK and should probably be removed at some point).

antoninbas:
I think it should be more satisfactory now

PR #470, opened 2019-05-14
Change locking model in DeviceMgr

antoninbas:
Before processing each WriteRequest batch, we now ensure that the thread
processing the batch has exclusive access to all the objects modified in
the batch. This means that the underlying target can safely assume that
individual P4 forwarding elements are not accessed by concurrent
threads. This probably has some non neglectable overhead as we need to
iterate over the entire batch before we start processing it and we have
to perform some set operations.

This change means that we can probably get rid of some other mutexes in
the implementation (e.g. the ActionProfMgr mutex), but this needs to be
done carefully. We may be able to change TableStore locking as well; one
issue is the idle-timeout thread that accesses the store to convert the
PI match key to a P4Runtime match key.

### DIFF
--- a/proto/frontend/Makefile.am
+++ b/proto/frontend/Makefile.am
@@ -12,6 +12,8 @@ AM_CPPFLAGS = \
 
 libpifeproto_la_SOURCES = \
 src/device_mgr.cpp \
+src/access_arbitration.h \
+src/access_arbitration.cpp \
 src/action_prof_mgr.h \
 src/action_prof_mgr.cpp \
 src/table_info_store.h \
--- a/proto/frontend/src/access_arbitration.cpp
+++ b/proto/frontend/src/access_arbitration.cpp
@@ -1,0 +1,214 @@
+/* Copyright 2019-present Barefoot Networks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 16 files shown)


jafingerhut:
Does the possible overhead seem worth considering an option on WriteRequest messages where the client can give a list of all tables/entities it will access in the rest of the message (e.g. just the id of 1 table if all writes are for 1 table, or N tables/counters/meters/etc. if it is a batch containing writes to all N of those entities).   If that optional part of the WriteRequest is present, the server locks only those entities, without first scanning through the entire WriteRequest batch.

If the server ever encounters a WriteRequest to any entity that wasn't in the new optional part of the message, it is automatically a failed request.

If the new part of the message was optional, it can be added in a backwards compatible fashion.  Only clients wanting the extra little bit of performance need to bother using it.

antoninbas:
It would be great to have this in my case, but I don't feel good letting the specifics of my particular implementation dictate what goes into the Protobuf interface definition. So we would need to check with other P4Runtime stakeholders. For example, it's perfectly legitimate for an implementation to serialize all the WriteRequests and then such information isn't required. I also considered doing that in my case, but I wanted the ability to perform small batches quickly even in the presence of a large ongoing batch. As I start working on the implementation of the `watch` attribute for action profile group members, I also wanted the ability to get exclusive access to a P4 object without blocking read responses or write requests to other objects.
Maybe this locking model will evolve again in the future, especially if we start supporting role partitioning.

I haven't quantified the overhead, but I would imagine it is not that important compared to the actual processing of the batch.

PR #471, opened 2019-05-16
Support for watch port attribute for ActionProfileGroup members

antoninbas:
Watch ports are handled entirely by the DeviceMgr. The PI layer is
responsible for generating notifications (through callbacks) when the
status of a DP port changes. The WatchPortEnforcer instance (owned by
the DeviceMgr) implements a task queue which will process the
notifications asynchronously and activate / deactivate action profile
group members as needed.

The notion of activated / deactivated member is new. A deactivated
member is still present in a group (and will show as part of the group
when fetching the list of members) but cannot be selected by the data
path for packet forwarding any more.

We introduce 2 new PI APIs, pi_act_prof_grp_activate_mbr and
pi_act_prof_grp_deactivate_mbr, which targets are required to implement,
and which are used by the WatchPortEnforcer to activate / deactivate
members as needed when receiving port status event notifications from
the target.

When using the individual add / remove APIs to manage groups, we assume
that members are always added in the "activated" state. When using the
set members API to manage groups, we provide a new function parameter
which is used to choose whether each member should be activated /
deactivated.

I briefly toyed with the idea of letting the target implement support
for watch port itself, but ultimately decided against it. Some targets
may have optimized ways of dealing with ports going down however, and if
this is the case we may consider introducing the ability to delegate
watch support to the target in the future.


jafingerhut:
I created a few related questions in this issue, which arose while reading this PR's description: https://github.com/p4lang/p4runtime/issues/203

PR #487, opened 2019-07-22
Remove gNMI dependency from pifeproto bazel target

Yi-Tseng:
Remove gNMI dependency since we are not using it in "pifepproro"


Yi-Tseng:
Yes, `bazel test //proto/tests:all` pass

PR #502, opened 2020-01-15
Cjson memory delete to use correct API

saynb:
* Making cJSON_free and cJSON_malloc in third_party/cJSON as public APIs
* For objects allocated with cJSON_Print or cJSON_PrintUnformatted,
use cJSON_free to deallocate them instead of using free directly


saynb:
@antoninbas When can this be merged in? Anything else needed from my side?

PR #504, opened 2020-02-02
Add support for optional match type

antoninbas:
V1model and P4Runtime now support the "optional" match type. The easiest
way for us to support it in this P4Runtime implementation (PI) is to
modify the pi_p4info_t struct to support "optional" natively, while
treating it as a "ternary" match when serializing the match key for the
PI backend. This ensures that:
 * we can properly validate the P4Runtime messages since we can
   distinguish between ternary / optional matches.
 * this information is also available to target implementations, in case
   it is required when writing entries to HW.
 * for targets which implement optional matches as ternary matches
   (where the mask is either all 0s or all 1s) very few changes are
   required.

The PI_ABI_VERSION number is increased to 2, since although the changes
required in most targets are minimal, this update is likely to break
existing PI backend libraries.


antoninbas:
This PR is tagged as a draft because it depends on https://github.com/p4lang/p4runtime/pull/259

jafingerhut:
I do not have the knowledge required to give a good review of these changes, sorry, but I am willing to test them out after they are merged in and look for issues in functionality.

I am only commenting to say that, plus that I think all of the prerequisites have been merged now.

PR #519, opened 2020-09-30
Support Python 3 in PI

fruffy:
Not much to change here. cpplint is the biggest change.

### DIFF
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,7 +1,0 @@
-set noparent
-filter=-build/include_dirname,-build/c++11
-filter=-build/include_subdir
-filter=-build/include_order
-filter=-build/include_what_you_use
-filter=-runtime/string
-filter=-runtime/explicit
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,0 +1,6 @@
+set noparent
+filter=-build/include_dirname,-build/c++11
+filter=-build/include_subdir
+filter=-build/include_order
+filter=-build/include_what_you_use
+filter=-runtime/string
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -4,3 +4,4 @@ filter=-build/include_subdir
 filter=-build/include_order
 filter=-build/include_what_you_use
 filter=-runtime/string
+filter=-runtime/explicit
--- a/tools/clang_format_check.py
+++ b/tools/clang_format_check.py
@@ -10,6 +10,7 @@ import subprocess
 import xml.etree.ElementTree as ET
 import sys
 import re
+import io
 
 from collections import namedtuple
 Replacement = namedtuple("Replacement", "offset length text")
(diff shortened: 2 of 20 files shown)


CPPLINT.cfg
```diff
@@ -0,0 +1,7 @@
+set noparent
+filter=-build/include_dirname,-build/c++11
+filter=-build/include_subdir
+filter=-build/include_order
+filter=-build/include_what_you_use
+filter=-runtime/string
+filter=-runtime/explicit
```
fruffy, line 7: I also had to add a root config with some ignore rules here.

antoninbas (approved):
LGTM

In the bmv2 PR, you created the following symlink for the Python interpreter in the Dockerfile:

```bash
ln -sf /usr/bin/python3 /usr/bin/python
```

Curious about why you are not doing the same thing here. Also curious about why you only did it for one of the bmv2 Dockerfiles and not all of them.

fruffy:
I wanted to avoid affecting downstream builds that do not yet support Python 3 with this. Now that all of them have support we can probably start forcing Python 3 the reverse way (p4c->bmv2->pi->third_party).

fruffy:
So I guess we have to wait until the `p4lang/behavioral-model:no-pi image` updates? 
No idea why clang-format is throwing that error maybe `for line in open(file, "r", encoding="utf-8"):` can help but I cannot reproduce this in the local image.

antoninbas:
You can probably add a log message to `tools/clang_format_check.py` to figure out which file is the issue

fruffy:
Okay this is ready for review now. Now all that is left is to migrate the cpplint.py scripts in p4c, PI, behavioral-model to Python 3...

tools/clang_format_check.py
```diff
@@ -106,7 +106,7 @@ def errors_from_replacements(file, replacements=[]):
 
     lines = [0]  # line index to character offset
     file_content = ""
-    for line in open(file, "r"):
+    for line in open(file, "r", encoding="utf-8"):
```
antoninbas, line 109: did you get to the bottom of this? which file was responsible for the failure and why this flag is required?

fruffy: https://github.com/p4lang/PI/blob/master/src/p4info/p4info_struct.c#L58

I think it is the section sign here. I can also remove that and do not add the encoding.

antoninbas: Adding the encoding is fine. I think ultimately it is because `open` means different things in Python 2 & 3. Speaking of which, if we still want to support both versions, should we use `io.open` here as suggested in https://docs.python.org/3/howto/pyporting.html ?

fruffy: So the way I understand this is that `io.open` has the same meaning in Python 2 that  `open` does in Python 3.
I just ran the Dockerfile with Python 2 instead of Python 3 and they all passed. Seems like open can also be used with that parameter in Python 2.  But no guarantees...

antoninbas (approved):
LGTM

Issue #533, opened 2021-02-15
Build fails, due to usage of ::stdcxx::shared_ptr instead of std::shared_ptr (Apache Thrift 0.13.0)


dreibh:
The build fails, due to usage of ::stdcxx::shared_ptr instead of std::shared_ptr (Apache Thrift 0.13.0):

conn_mgr.cpp:48:36: error: field ‘transport’ has incomplete type ‘boost::shared_ptr<apache::thrift::transport::TTransport>’
   48 |   ::stdcxx::shared_ptr<TTransport> transport{nullptr};
      |                                    ^~~~~~~~~
In file included from /usr/include/boost/throw_exception.hpp:36,
                 from /usr/include/boost/numeric/conversion/converter_policies.hpp:17,
                 from /usr/include/boost/numeric/conversion/converter.hpp:14,
                 from /usr/include/boost/numeric/conversion/cast.hpp:33,
                 from /usr/include/thrift/transport/TTransportException.h:23,
                 from /usr/include/thrift/transport/TTransport.h:24,
                 from /usr/include/thrift/protocol/TProtocol.h:28,
                 from /usr/include/thrift/TProcessor.h:24,
                 from /usr/include/thrift/TDispatchProcessor.h:22,
                 from /usr/include/c++/10/bm/SimplePreLAG.h:10,
                 from conn_mgr.h:24,
                 from conn_mgr.cpp:21:

Setup:
* Ubuntu 20.04
* Apache Thrift 0.13.0, compiled from sources (tag v0.13.0)

Configure options:
./configure --prefix=/usr --with-proto --with-fe-cpp --with-cli --with-internal-rpc --with-gnu-ld --with-bmv2 --includedir=/usr/include/c++/10

dreibh:
Quick work-around:
`sed -e "s#::stdcxx::shared_ptr#std::shared_ptr#g"  -i targets/bmv2/conn_mgr.cpp`

I am going to provide a proper pull request with the fix.

dreibh:
The pull request https://github.com/p4lang/PI/pull/534 fixes the issue.

PR #534, opened 2021-02-15
Fix for build issue #533

dreibh:
Fix for https://github.com/p4lang/PI/issues/533:
The build failed, due to usage of ::stdcxx::shared_ptr instead of std::shared_ptr (Apache Thrift 0.13.0). This pull request simply fixes the issue by replacing ::stdcxx::shared_ptr with std::shared_ptr.

### DIFF
--- a/targets/bmv2/conn_mgr.cpp
+++ b/targets/bmv2/conn_mgr.cpp
@@ -45,7 +45,7 @@ using namespace ::apache::thrift::transport;  // NOLINT(build/namespaces)
 namespace pibmv2 {
 
 struct ClientImp {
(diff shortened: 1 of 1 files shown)


antoninbas (approved):
I would have preferred a solution like this one to ensure backwards compatibility with older thrift versions: https://github.com/p4lang/behavioral-model/commit/e90ae599cae840dcbca27aad4e393fdaed780b2b

However, it seems that this solution works fine with Thrift 0.11 (otherwise I would expect the CI build to fail), which is officially the oldest version we support for p4lang software.

PR #539, opened 2021-03-16
Remove ppa:ubuntu-toolchain-r/test from Dockerfiles

antoninbas:
For some reason, having this PPA makes it impossible to install clang,
it creates a "dependency hell". This is recent a issue and wasn't
present a couple weeks ago. To avoid the issue altogether, rather than
waste a lot of time debugging it, we remove the PPA, which means we no
longer run a CI job with a "recent" version of gcc (gcc 7). We still
test with the default version of gcc for Ubuntu 16.04, along with
clang-3.8 and clang-8 (updated from clang-6.0).

The long-term solution should really be to update the base image from
Ubuntu 16.04 to 20.04 (https://github.com/p4lang/third-party/issues/25).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,10 @@ matrix:
   include:
     - name: gcc
       env: CXX=g++ CC=gcc
-    - name: gcc-7
-      env: CXX=g++-7 CC=gcc-7
     - name: clang-3.8
       env: CXX=clang++-3.8 CC=clang-3.8
-    - name: clang-6.0
-      env: CXX=clang++-6.0 CC=clang-6.0
+    - name: clang-8
+      env: CXX=clang++-8 CC=clang-8
 
 install:
   - docker build -t pi --build-arg IMAGE_TYPE=test --build-arg CC=$CC --build-arg CXX=$CXX -f Dockerfile.bmv2 .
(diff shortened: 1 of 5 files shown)


mbudiu-vmw:
@antoninbas : all our CI pipelines for p4c are now failing to install the software. @fruffy suggests that this change is responsible.

antoninbas:
@mbudiu-vmw @fruffy

It seems that p4c uses the bmv2 Docker image as the base image: https://github.com/p4lang/p4c/blob/192d128c48d0a266f7610566486acf9074e6a52a/Dockerfile#L1.

You can see in Dockerhub that the bmv2 image has not been updated in the last 3 months: https://hub.docker.com/repository/docker/p4lang/behavioral-model

So I think you need to look elsewhere for an explanation.

I have been experiencing build failures as well, which is what I am trying to solve with this PR. I have a similar PR for bmv2 (https://github.com/p4lang/behavioral-model/pull/984) which I intend to merge today. Maybe when this is merged, it will solve your build issues as well. I have determined that - at least on my case - the issues were caused by the test toolchain PPA, so I am no longer adding it to apt sources. Since the PPA was added in the base image, it would also be present when building the p4c Docker image, so maybe the root cause for your build failures is the same.

antoninbas:
Note that as soon as I merge the bmv2 PR, the bmv2 Docker image will be updated. But at the time of writing this, it has indeed not been updated in 3 months.

PR #541, opened 2021-03-26
Collected build fixes

dreibh:
Collected build fixes:
- Compile with -Wno-deprecated-declarations for Ubuntu 20.10 and 21.04.
- Ensure that all files get included into "make dist" tarball.
- Added missing include of stdint.h.
- Fix for compiling with Apache Thrift 0.13.
- Fixed typo.


targets/bmv2/conn_mgr.cpp
```diff
@@ -45,7 +45,7 @@ using namespace ::apache::thrift::transport;  // NOLINT(build/namespaces)
 namespace pibmv2 {
```
jafingerhut, line 45: The changes recommended in this file were made via some other PR already.

PR #546, opened 2021-04-20
Unsafe proto release cleanup

mkruskal-google:
MUST_USE_RESULT is being added to these release methods, which will break this code as it's currently written.  This change is part of an internal google effort to fix these instances before enabling PROTOBUF_MUST_USE_RESULT.

The unsafe arena variants of set_allocated_* and release_* will not require the result to be checked, and will also avoid unwanted copies or memory leaks if the protos are owned by an arena.

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -3,14 +3,15 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("//bazel:workspace_rule.bzl", "remote_workspace")
 
-GNMI_COMMIT = "39cb2fffed5c9a84970bde47b3d39c8c716dc17a";
-GNMI_SHA = "3701005f28044065608322c179625c8898beadb80c89096b3d8aae1fbac15108";
+GNMI_COMMIT = "39cb2fffed5c9a84970bde47b3d39c8c716dc17a"
+GNMI_SHA = "3701005f28044065608322c179625c8898beadb80c89096b3d8aae1fbac15108"
+
 # P4RUNTIME_TAG = "1.3.0"
 # We cannot use the latest release (v1.3.0) as we need to include a recent fix
 # to support Bazel 4.0. More precisely, this fix updates the Bazel build
 # dependencies to more recent versions compatible with Bazel 4.0.
-P4RUNTIME_COMMIT = "0d40261b67283999bf0f03bd6b40b5374c7aebd0"
-P4RUNTIME_SHA="d1b31378f58fa7c6039edac06299acfe77ec130e4dffb4704b4e5a30293bd2a5"
+P4RUNTIME_COMMIT = "e9c0d196c4c2acd6f1bd3439f5b30b423ef90c95"
+P4RUNTIME_SHA = "039fb488af8b5ae8f2a5dd05ca4385b1ab33099165c4d61405fb50eedf0c6324"
 
 def PI_deps():
     """Loads dependencies needed to compile PI."""
@@ -50,7 +51,7 @@ def PI_deps():
             build_file = "@//bazel:external/gnmi.BUILD",
             patch_cmds = [
                 "sed -i.bak 's#github.com/openconfig/gnmi/proto/##g' gnmi/gnmi.proto",
-                "rm gnmi/gnmi.proto.bak"
+                "rm gnmi/gnmi.proto.bak",
             ],
         )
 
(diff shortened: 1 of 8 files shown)


antoninbas (commented):
Thanks for the patch. Is there any public resource that you can link to about that upcoming Protobuf change?
I believe some tests under https://github.com/p4lang/PI/tree/main/proto/tests also use these functions.

fowles:
The current usage is actually out of contract for protos that are on arenas.  The internal changes to google are to make broader use of arenas internally.

mkruskal-google:
Ah, thanks for the pointer!  I don't think we build the tests internally so I didn't pick them up.

See Matt's comment about the changes.  This shouldn't really have any affect on you, being outside google, but this does clean up a misuse that could cause issues for you in the future

antoninbas (approved):
LGTM

antoninbas:
@mkruskal-google CI is failing because of cpplint failures
```
PI/tools/../proto/tests/test_proto_fe.cpp:2685:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
PI/tools/../proto/tests/test_proto_fe.cpp:2697:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
```

mkruskal-google:
Thanks Antonin, fixed!

antoninbas:
I see some build failures:

```
src/device_mgr.cpp:605:17: error: no member named 'unsafe_arena_set_allocated_entity' in 'p4::v1::Update'
        update->unsafe_arena_set_allocated_entity(&entity);
        ~~~~~~  ^
src/device_mgr.cpp:609:16: error: no member named 'unsafe_arena_release_entity' in 'p4::v1::Update'
        update.unsafe_arena_release_entity();
        ~~~~~~ ^
src/device_mgr.cpp:1601:9: error: no member named 'unsafe_arena_set_allocated_error' in 'p4::v1::StreamMessageResponse'
    msg.unsafe_arena_set_allocated_error(&stream_error);
    ~~~ ^
src/device_mgr.cpp:1603:9: error: no member named 'unsafe_arena_release_error' in 'p4::v1::StreamMessageResponse'
    msg.unsafe_arena_release_error();
    ~~~ ^
```

The CI uses Protobuf / protoc 3.6. Is it possible that these new methods were introduced in a more recent version?

mkruskal-google:
Yep, I'm looking into that now, but I came to the same conclusion.  I see two options:
1) Bump up the version of protobuf used for CI
2) Find a workaround (e.g. "option cc_enable_arenas = true" should enable it in some past versions, although I'm not sure about the one you're using)

I'm looking into 2, but WDYT of 1?

antoninbas:
I would personally be ok with updating the Protobuf version used for CI, but it may impact other people so it may be a bit tedious to do. We use a single version for all p4.org projects, and some downstream projects user this version as well. It comes from https://github.com/p4lang/third-party.

I guess one solution would be to use one method or the other based on the value of the `GOOGLE_PROTOBUF_VERSION` preprocessor macro.
It's not perfect because there can be a discrepancy between the protoc version and the version of the protobuf headers. However, there is the following guarantee: `protoc version <= protobuf version`, so I think it will work for your scenario and for CI as well. Only issue would be if someone has a version mismatch, with 1) a version of protoc that doesn't generate the new methods and 2) a version of Protobuf that matches to a version of protoc that does generate them. Personally I can live with that.

Have you considered just fixing the usage of `set_allocated_*` and `release_*` in the PI code?

mkruskal-google:
https://github.com/p4lang/p4runtime/pull/348 *should* fix this issue without bumping up the version.

I did consider removing the usage.  The problem is this is a perfectly valid use case.  You're adding an externally owned pointer to the proto, serializing it, and then removing it again to avoid freeing it.  This is the best way to avoid a copy.

mkruskal-google:
Do I need to do anything to pull in the changes to p4runtime and rerun the tests?

antoninbas:
Yes you need to update the Git submodule pointer in that folder: https://github.com/p4lang/PI/tree/main/proto
And update the P4Runtime commit for the Bazel build: https://github.com/p4lang/PI/blob/main/bazel/deps.bzl

mkruskal-google:
Thanks Antonin!  I *think* that should cover it?  I'm a little nervous about that comment in deps.bzl

antoninbas (approved):

PR #566, opened 2021-12-21
Use function macros for PI logging stubs

### LINKED ISSUE #565: Fix compile warning without PI_LOG_ON
https://github.com/p4lang/PI/blob/66da143288d4f02f17cd6a63d0f4b2a87b011499/src/utils/logging.h#L28-L36

To work as intended, the empty macros above need to be function macros instead of object macros, i.e.,
```C
#define FOO(...)
```
instead of
```C
#define FOO
```

antoninbas:
Fixes #565


rst0git (approved):
LGTM

PR #574, opened 2022-02-07
Updates for Ubuntu 20.04

antoninbas:


Dockerfile.bmv2
```diff
@@ -16,14 +16,14 @@ ARG MAKEFLAGS=-j2
...
+# We install the default version of GCC (GCC 9), as well as clang 8 and clang 10.
 ARG CC=gcc
 ARG CXX=g++
 
 ENV PI_DEPS automake \
             build-essential \
-            clang-3.8 \
             clang-8 \
+            clang-10 \
             clang-format-8 \
```
rst0git, line 27: > Can't run 'clang-format', please make sure it is installed 

The [`clang-format-10`](https://packages.ubuntu.com/focal/clang-format-10) package might fix this error

Dockerfile.bmv2
```diff
@@ -1,5 +1,5 @@
 FROM p4lang/behavioral-model:no-pi
```
rst0git, line 1: Do we need to update the `p4lang/behavioral-model:no-pi` image as well?

PR #577, opened 2022-03-09
Replace Travis job with Github actions

antoninbas:


.github/workflows/test.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: Check whether tests need to be run based on diff
+    runs-on: [ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - uses: antrea-io/has-changes@v1
+      id: check_diff
+      with:
+        args: docs/* *.md bazel/*
```
rst0git, line 19: @antoninbas GitHub Actions supports this functionality with [`paths` and `paths-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

antoninbas: Last time I used `paths-ignore`, it didn't support returning a passing status when skipping the job.
I don't think this has changed. https://stackoverflow.com/questions/66751567/return-passing-status-on-github-workflow-when-using-paths-ignore

PR #588, opened 2023-04-11
Increase the max value for the clone session ID

fruffy:
PI restricts the maximum clone session ID to 512, but this is a rather conservative number. Some P4 programs may use the full range of the `pi_clone_session_id_t` (which appears to be uint16_t) for the behavioral model. Increase the upper bound to 65535.

### DIFF
--- a/proto/frontend/src/pre_clone_mgr.h
+++ b/proto/frontend/src/pre_clone_mgr.h
@@ -29,6 +29,7 @@
 
 #include "google/rpc/status.pb.h"
 #include "p4/v1/p4runtime.pb.h"
+
 #include "pre_mc_mgr.h"
 
 namespace pi {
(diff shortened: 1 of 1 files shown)


antoninbas (commented):
I think it may be a better idea to set this to 32768, for the following reasons:
- this is the max value supported by simple_switch, which reserves the left-most bit (https://github.com/p4lang/behavioral-model/blob/e9fa7dc687f334e5cf327e0c993fc1a351d224c0/targets/simple_switch/register_access.h#L46). There will be a better error message if the error is caught early in PI.
- for clone sessions to more than 1 port, PI will program a multicast group. IIRC, there may only be 32768 mgids available for this usage (with another 32768 reserved for user-defined multicast groups).

The comment could also be updated to indicate that 32768 is the max value supported by bmv2 simple_switch_grpc, but that different targets may support a different number of clone sessions. (there is also a typo in the comment: s/te/the).

antoninbas (approved):

antoninbas (approved):

PR #598, opened 2023-07-28
Add support for replicas with new port kind "port".

smolkaj:
See https://github.com/p4lang/p4runtime/pull/331 for the corresponding change in P4Runtime.

I have tested this internally at Google and confirmed that it works. I did not know how to write (unit?) test for this project.

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -21,7 +21,7 @@ def PI_deps():
             name = "com_github_p4lang_p4runtime",
             # urls = ["https://github.com/p4lang/p4runtime/archive/v%s.zip" % P4RUNTIME_TAG],
             urls = ["https://github.com/p4lang/p4runtime/archive/%s.zip" % P4RUNTIME_COMMIT],
-            sha256 = P4RUNTIME_SHA,
+            # sha256 = P4RUNTIME_SHA,
             # strip_prefix = "p4runtime-%s/proto" % P4RUNTIME_TAG,
             strip_prefix = "p4runtime-%s/proto" % P4RUNTIME_COMMIT,
         )
@@ -62,6 +62,7 @@ def PI_deps():
             name = "com_google_googleapis",
             remote = "https://github.com/googleapis/googleapis",
             commit = "9fe00a1330817b5ce00919bf2861cd8a9cea1a00",
+            sha256 = "b35510fd77d19562893da449b41d6fb44d2b9f391bd7928208984569e95dbe15",
         )
 
     if "com_github_nelhage_rules_boost" not in native.existing_rules():
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -11,7 +11,7 @@ GNMI_SHA="3701005f28044065608322c179625c8898beadb80c89096b3d8aae1fbac15108";
 # to support Bazel 5.0. More precisely, this fix updates the Bazel build
 # dependencies to more recent versions compatible with Bazel 5.0.
 P4RUNTIME_COMMIT="d76a3640a223f47a43dc34e5565b72e43796ba57"
-P4RUNTIME_SHA="5e418c86b6a377da0f3f92c1ae34359d8cd064f26b2b74ec128c6fbefc414173"
+P4RUNTIME_SHA="944648d331b5645d490ab592d2478d0d55bb7442edc068938efd79937af99d1d"
 
 def PI_deps():
     """Loads dependencies needed to compile PI."""
@@ -21,7 +21,7 @@ def PI_deps():
             name = "com_github_p4lang_p4runtime",
             # urls = ["https://github.com/p4lang/p4runtime/archive/v%s.zip" % P4RUNTIME_TAG],
             urls = ["https://github.com/p4lang/p4runtime/archive/%s.zip" % P4RUNTIME_COMMIT],
-            # sha256 = P4RUNTIME_SHA,
+            sha256 = P4RUNTIME_SHA,
             # strip_prefix = "p4runtime-%s/proto" % P4RUNTIME_TAG,
             strip_prefix = "p4runtime-%s/proto" % P4RUNTIME_COMMIT,
         )
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -62,7 +62,6 @@ def PI_deps():
             name = "com_google_googleapis",
             remote = "https://github.com/googleapis/googleapis",
             commit = "9fe00a1330817b5ce00919bf2861cd8a9cea1a00",
-            sha256 = "b35510fd77d19562893da449b41d6fb44d2b9f391bd7928208984569e95dbe15",
         )
 
     if "com_github_nelhage_rules_boost" not in native.existing_rules():
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -10,7 +10,7 @@ GNMI_SHA="3701005f28044065608322c179625c8898beadb80c89096b3d8aae1fbac15108";
 # We cannot use the latest release (v1.3.0) as we need to include a recent fix
 # to support Bazel 5.0. More precisely, this fix updates the Bazel build
 # dependencies to more recent versions compatible with Bazel 5.0.
-P4RUNTIME_COMMIT="e5d8c89e81ef655f79dbdd6749623fdc764ee90f"
+P4RUNTIME_COMMIT="d76a3640a223f47a43dc34e5565b72e43796ba57"
 P4RUNTIME_SHA="5e418c86b6a377da0f3f92c1ae34359d8cd064f26b2b74ec128c6fbefc414173"
 
 def PI_deps():
--- a/proto/frontend/src/action_prof_mgr.cpp
+++ b/proto/frontend/src/action_prof_mgr.cpp
@@ -63,10 +63,8 @@ pi_port_t watch_port_p4rt_to_pi(int watch) {
 }
 
 pi_port_t watch_port_p4rt_to_pi(const std::string &watch) {
-  if (watch == "") return WatchPortEnforcer::INVALID_WATCH;
   pi_port_t v;
-  Status status = common::bytestring_to_pi_port(watch, &v);
-  if (!IS_OK(status) || v < 0 || v == p4v1::SDN_PORT_UNKNOWN)
+  if (!IS_OK(common::bytestring_to_pi_port(watch, &v)))
       return WatchPortEnforcer::INVALID_WATCH;
   return v;
 }
--- a/proto/frontend/src/action_prof_mgr.cpp
+++ b/proto/frontend/src/action_prof_mgr.cpp
@@ -57,18 +57,6 @@ using OneShotMember = ActionProfAccessOneshot::OneShotMember;
 
 namespace {
 
-// temporary until we have port translation support
-
-int bytestring_to_integer(const std::string &str) {
-  int v = 0;
-  if (str.size() > 4) return -1;
-  for (auto c : str) {
-    v = v << 8;
-    v += static_cast<int>(static_cast<unsigned char>(c));
-  }
-  return v;
-}
-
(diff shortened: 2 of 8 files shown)


smolkaj:
@antoninbas looks like there is a test failure in `test_proto_fe`, but I don't see an error message in the GitHub CI run. Would you happen to know how I can go about debugging this?

smolkaj:
+ @fruffy on the off chance that you can help me out here :)

proto/frontend/src/pre_mc_mgr.cpp
```diff
@@ -42,6 +45,17 @@ using Code = ::google::rpc::Code;
...
+std::tuple<p4v1::Replica::PortKindCase, pi_mc_port_t, size_t> ReplicaPortAsTuple(
+    const ReplicaPort &port) {
+  return std::make_tuple(port.port_kind, port.port_id, port.num_bytes);
+}
+bool operator==(const ReplicaPort &x, const ReplicaPort &y) {
+  return ReplicaPortAsTuple(x) == ReplicaPortAsTuple(y);
+}
+bool operator<(const ReplicaPort &x, const ReplicaPort &y) {
+  return ReplicaPortAsTuple(x) < ReplicaPortAsTuple(y);
+}
```
antoninbas, line 57: not sure I see the value of building a tuple here as opposed to comparing individual fields?

proto/frontend/src/pre_mc_mgr.cpp
```diff
@@ -147,13 +161,69 @@ struct PreMcMgr::NodeCleanupTask : public McLocalCleanupIface {
...
+        RETURN_ERROR_STATUS(Code::UNIMPLEMENTED,
+                            "Multicast group replica `port` has %d bytes, but "
+                            "only %d bytes are supported",
+                            egress_port.num_bytes, sizeof(pi_mc_port_t));
+      }
+      egress_port.port_id = 0;
+      for (char byte : replica.port()) {
+        egress_port.port_id <<= 8;
+        egress_port.port_id += static_cast<uint8_t>(byte);
+      }
```
antoninbas, line 185: could you reuse the existing function: https://github.com/p4lang/PI/blob/f043e6f5f4271076ad7e58aa9889c82dbfc8c3ca/proto/frontend/src/action_prof_mgr.cpp#L62. It probably makes sense to move it to common.h

bazel/deps.bzl
```diff
@@ -21,7 +21,7 @@ def PI_deps():
             name = "com_github_p4lang_p4runtime",
             # urls = ["https://github.com/p4lang/p4runtime/archive/v%s.zip" % P4RUNTIME_TAG],
             urls = ["https://github.com/p4lang/p4runtime/archive/%s.zip" % P4RUNTIME_COMMIT],
-            sha256 = P4RUNTIME_SHA,
+            # sha256 = P4RUNTIME_SHA,
```
antoninbas, line 24: will commenting this out always give a warning during bazel builds?
I think if you omit it and build locally, Bazel will give you the correct sha to use

proto/frontend/src/pre_mc_mgr.cpp
```diff
@@ -147,13 +161,69 @@ struct PreMcMgr::NodeCleanupTask : public McLocalCleanupIface {
...
+          bytes[i] = value & 0xffu;
+          value >>= 8;
+        }
+      }
+      RETURN_OK_STATUS();
+    }
+    default:
+      RETURN_ERROR_STATUS(
+          Code::INTERNAL,
+          "Unset `port_kind` in intenral `ReplicaPort` representation");
```
antoninbas, line 215: typo: s/intenral/internal

antoninbas (commented):
for the test failure, you can probably reproduce locally. If not, you could edit https://github.com/p4lang/PI/blob/f043e6f5f4271076ad7e58aa9889c82dbfc8c3ca/.github/workflows/test.yml#L51 to output the test results on failure: `make check -j2 || cat ./test-suite.log`

I think it used to print test results on failure, but maybe something changed at some point

It would make sense to add a gtest, It can be done by:
- duplicating the following function as `push_back_new` and using the new field: https://github.com/p4lang/PI/blob/f043e6f5f4271076ad7e58aa9889c82dbfc8c3ca/proto/tests/test_proto_fe.cpp#L3510
- adding a unit test in the `PREMulticastTest` group, which uses this new version. It can be a simplified copy of `TEST_F(PREMulticastTest, Write)`

proto/frontend/src/pre_mc_mgr.cpp
smolkaj: My motivation was simplicity/readability.
The `<` operator becomes a bit convoluted:
```
bool operator<(const ReplicaPort &x, const ReplicaPort &y) {
  if (x.port_kind < y.port_kind) return true;
  if (y.port_kind < x.port_kind) return false;
  if (x.port_id < y.port_id) return true;
  if (y.port_id < x.port_id) return false;
  return x.num_bytes < y.num_bytes;
}
```
Happy to change it if you prefer the above.

smolkaj:
Thanks for the review & tips! Will get back with an updated PR shortly.

bazel/deps.bzl
smolkaj: Fixed.

proto/frontend/src/pre_mc_mgr.cpp
```diff
@@ -172,21 +242,29 @@ PreMcMgr::read_group(
...
-      replica->set_egress_port(static_cast<uint32_t>(port));
-      replica->set_instance(rid);
+      auto &replica = *group_entry->add_replicas();
+      SetReplicaPort(port, replica);
+      replica.set_instance(rid);
     }
   }
 }
 
+static std::vector<pi_mc_port_t> GetPiEgressPorts(
```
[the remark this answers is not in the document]
smolkaj: Done (here and for the other functions I added).

proto/frontend/src/pre_mc_mgr.cpp
smolkaj: Fixed.

proto/frontend/src/pre_mc_mgr.cpp
smolkaj: I moved this logic and the reverse conversion into common.h, but ended up modifying it a bit to make it more robust & reusable.

smolkaj:
I was able to extract the error message, but don't understand why it's failing yet:
```
[ RUN      ] PREMulticastTest.Write
unknown file: Failure

Unexpected mock function call - taking default action specified at:
mock_switch.cpp:1291:
    Function call: mc_node_modify(0, @0x7ffe0f7aa980 { 2 })
          Returns: 0
Google Mock tried the following 1 expectation, but it didn't match:

test_proto_fe.cpp:3621: EXPECT_CALL(*mock, mc_node_modify(_, ElementsAre(port1, port3)))...
  Expected arg #1: has 2 elements where
element #0 is equal to 1,
element #1 is equal to 3
           Actual: { 2 }, which has 1 element
         Expected: to be called once
           Actual: never called - unsatisfied and active
unknown file: Failure

Unexpected mock function call - taking default action specified at:
mock_switch.cpp:1291:
    Function call: mc_node_modify(0, @0x7ffe0f7aa980 { 2 })
          Returns: 0
Google Mock tried the following 1 expectation, but it didn't match:

test_proto_fe.cpp:3[621](https://github.com/p4lang/PI/actions/runs/5695793632/job/15439664520?pr=598#step:4:622): EXPECT_CALL(*mock, mc_node_modify(_, ElementsAre(port1, port3)))...
  Expected arg #1: has 2 elements where
element #0 is equal to 1,
element #1 is equal to 3
           Actual: { 2 }, which has 1 element
         Expected: to be called once
           Actual: called once - saturated and active
test_proto_fe.cpp:3621: Failure
Mock function called more times than expected - taking default action specified at:
mock_switch.cpp:1291:
    Function call: mc_node_modify(1, @0x7ffe0f7aa980 { 1, 3 })
          Returns: 0
         Expected: to be called once
           Actual: called twice - over-saturated and active
[  FAILED  ] PREMulticastTest.Write (1 ms)
```
Interestingly, the new test I added passes.

smolkaj:
By the way, I'm not able to build things the non-bazel way from scratch to reproduce the issue...unfortunately I have very little experience with building stuff the autoconf way.

smolkaj:
All checks passing, this is ready for another look when you get a chance, @antoninbas.

proto/frontend/src/action_prof_mgr.cpp
```diff
@@ -57,29 +57,18 @@ using OneShotMember = ActionProfAccessOneshot::OneShotMember;
...
       WatchPortEnforcer::INVALID_WATCH : static_cast<pi_port_t>(watch);
 }
 
 pi_port_t watch_port_p4rt_to_pi(const std::string &watch) {
   if (watch == "") return WatchPortEnforcer::INVALID_WATCH;
-  auto v = bytestring_to_integer(watch);
-  if (v < 0 || v == p4v1::SDN_PORT_UNKNOWN)
+  pi_port_t v;
+  Status status = common::bytestring_to_pi_port(watch, &v);
+  if (!IS_OK(status) || v < 0 || v == p4v1::SDN_PORT_UNKNOWN)
```
antoninbas, line 69: The `v < 0` check doesn't seem meaningful anymore now that `bytestring_to_pi_port` returns a `Status`?

antoninbas (commented):
you checked in a zip file by mistake, please remove it

smolkaj:
> you checked in a zip file by mistake, please remove it

Done.

proto/frontend/src/action_prof_mgr.cpp
smolkaj: `bytestring_to_pi_port` allows arbitrary byte strings of size <= 4, essentially any `uint32_t`. That decision was loosely based on this part of the spec

image

which includes SDN port numbers where the highest bit is set.

Perhaps we want to be more restrictive and pull all these checks into `bytestring_to_pi_port`? I just wasn't sure what the right restrictions were. Happy to take suggestions here.

antoninbas: It may make sense to have these checks in `bytestring_to_pi_port` then, because `GetReplicaPort` also calls `bytestring_to_pi_port` and does not validate the value right now (only the returned status).

smolkaj: Done.

proto/frontend/src/common.cpp
```diff
@@ -99,6 +99,44 @@ std::string bytestring_pi_to_p4rt(const char *str, size_t n) {
...
+    value += static_cast<uint8_t>(byte);
+  }
+  // Casting from unsigned to signed is implementation-defined if the source
+  // value would not fit in the destination, so we use memcpy as a work-around.
+  static_assert(sizeof(pi_port_t) == 4, "invariant broken");
+  memcpy(result, &value, 4);
+
+  if (*result <= 0) {
+    RETURN_ERROR_STATUS(Code::INVALID_ARGUMENT,"Port must be non-negative.");
+  }
```
antoninbas, line 124: so you don't think we need to check for `SDN_PORT_UNKNOWN` here, or at the caller?

smolkaj: `SDN_PORT_UNKNOWN == 0`, so technically we are checking.

If we are okay with pulling in the P4RT dependency here, I can make this explicit?

antoninbas: current code lgtm

antoninbas (approved):

antoninbas:
@smolkaj can you fix the formatting in common.cpp so I can merge?

proto/frontend/src/common.cpp
smolkaj: SG

smolkaj:
Style should be fixed. Thanks for the patient review, @antoninbas! Highly appreciated.

PR #600, opened 2023-08-03
ptf: update stringify for compatibility with Py3

rst0git:
The [`decode('hex')`](https://docs.python.org/2.7/library/codecs.html#python-specific-encodings) method has been deprecated and removed in Python 3. Equivalent functionality in Python 3 is provided by [`int.to_bytes`](https://docs.python.org/3/library/stdtypes.html#int.to_bytes).

### DIFF
--- a/proto/ptf/base_test.py
+++ b/proto/ptf/base_test.py
@@ -78,11 +78,12 @@ def ipv4_to_binary(addr):
     to a string with binary contents expected by the Python P4Runtime
     client operations."""
     bytes_ = [int(b, 10) for b in addr.split('.')]
-    assert len(bytes_) == 4
-    # Note: The chr(b) call below will throw exception if any b is
-    # outside of the range [0, 255]], so no need to add a separate
-    # check for that here.
-    return "".join(chr(b) for b in bytes_)
+    if len(bytes_) != 4:
+        raise ValueError("Invalid IPv4 address format")
+    for b in bytes_:
+        if b < 0 or b > 255:
+            raise ValueError("IPv4 address contains out-of-range value")
+    return bytes(bytes_)
 
 def mac_to_binary(addr):
     """Take an argument 'addr' containing an Ethernet MAC address written
@@ -701,6 +702,28 @@ class TestStringify(unittest.TestCase):
         self.assertEqual(stringify(255), b'\xff')
         self.assertEqual(stringify(512), b'\x02\x00')
         self.assertEqual(stringify(1000), b'\x03\xe8')
+
+
+class TestIPv4ToBinary(unittest.TestCase):
+    def test_valid_ipv4(self):
+        """Test valid IPv4 addresses"""
+        self.assertEqual(ipv4_to_binary('0.0.0.0'), b'\x00\x00\x00\x00')
+        self.assertEqual(ipv4_to_binary('192.168.0.1'), b'\xc0\xa8\x00\x01')
+        self.assertEqual(ipv4_to_binary('255.255.255.255'), b'\xff\xff\xff\xff')
+
+    def test_invalid_ipv4(self):
+        """Test invalid IPv4 addresses (less or more than 4 bytes)"""
+        with self.assertRaises(ValueError):
+            ipv4_to_binary('10.0.1')
+        with self.assertRaises(ValueError):
+            ipv4_to_binary('10.0.1.2.3')
+
+    def test_out_of_range(self):
+        """Test IPv4 address with out-of-range values (above 255)"""
+        with self.assertRaises(ValueError):
+            ipv4_to_binary('256.0.0.0')
+        with self.assertRaises(ValueError):
+            ipv4_to_binary('10.1.300.4')
 
 
 if __name__ == '__main__':
--- a/proto/ptf/base_test.py
+++ b/proto/ptf/base_test.py
@@ -25,6 +25,7 @@ import sys
 import threading
 import time
 import queue
+import unittest
 
 import ptf
 from ptf.base_tests import BaseTest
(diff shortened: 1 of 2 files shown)


proto/ptf/base_test.py
```diff
@@ -63,9 +60,7 @@ def stringify(n, length):
     'length' bytes, it is represented in the fewest number of bytes it
     does fit into without loss of precision.  It always returns a
     string at least one byte long, even if value=width=0."""
-    h = '%x' % n
-    s = ('0'*(len(h) % 2) + h).zfill(length*2).decode('hex')
-    return s
+    return int.to_bytes(n, length=length)
```
antoninbas, line 63: only very recent versions of Python 3 (>= 3.11) have a default value for the `byteorder` argument.

proto/ptf/base_test.py
```diff
@@ -49,12 +49,9 @@ def __get__(self, instance, owner):
                        *(self.args or ()), **(self.keywords or {}))
 
 # Convert integer (with length) to binary byte string
-# Equivalent to Python 3.2 int.to_bytes
-# See
-# https://stackoverflow.com/questions/16022556/has-python-3-to-bytes-been-back-ported-to-python-2-7
 # TODO: When P4Runtime implementation is ready for it, use
 # minimum-length byte sequences to represent integers.  For unsigned
-# integers, this should only require removing the zfill() call below.
+# integers, this should only require removing the length call below.
```
antoninbas, line 54: I don't think this is a valid comment. There is no "length call"

proto/ptf/base_test.py
```diff
@@ -63,9 +60,7 @@ def stringify(n, length):
     'length' bytes, it is represented in the fewest number of bytes it
     does fit into without loss of precision.  It always returns a
     string at least one byte long, even if value=width=0."""
-    h = '%x' % n
-    s = ('0'*(len(h) % 2) + h).zfill(length*2).decode('hex')
-    return s
+    return int.to_bytes(n, length=length)
 
 def ipv4_to_binary(addr):
```
antoninbas, line 65: I don't think the implementations of `ipv4_to_binary` and `mac_to_binary` are valid either

maybe a simple unit test file would be beneficial to validate these 3 functions, maybe using Python 3.8

rst0git: I've updated the pull request with unit tests for `stringify`, `ipv4_to_binary` and `mac_to_binary`.
The unit tests are running in `p4lang/third-party:latest` container that provides Python 3.8.

proto/ptf/base_test.py
```diff
@@ -63,20 +61,19 @@ def stringify(n, length):
...
-    assert len(bytes_) == 4
-    # Note: The chr(b) call below will throw exception if any b is
-    # outside of the range [0, 255]], so no need to add a separate
-    # check for that here.
+    if len(bytes_) != 4:
+        raise ValueError("Invalid IPv4 address format")
+    for b in bytes_:
+        if b < 0 or b > 255:
+            raise ValueError("IPv4 address contains out-of-range value")
     return "".join(chr(b) for b in bytes_)
```
antoninbas, line 77: this returns a `str`, while `stringify` returns a `bytes` object
for correctness and consistency, I believe this should also return a `bytes` object, which can easily be done with `bytes(bytes_)`, since the `bytes` constructor can take as input an array of integers

same for `mac_to_binary` below

proto/ptf/base_test.py
```diff
@@ -49,12 +50,9 @@ def __get__(self, instance, owner):
                        *(self.args or ()), **(self.keywords or {}))
 
 # Convert integer (with length) to binary byte string
-# Equivalent to Python 3.2 int.to_bytes
-# See
-# https://stackoverflow.com/questions/16022556/has-python-3-to-bytes-been-back-ported-to-python-2-7
 # TODO: When P4Runtime implementation is ready for it, use
 # minimum-length byte sequences to represent integers.  For unsigned
-# integers, this should only require removing the zfill() call below.
+# integers, this should only require removing the `length` below.
```
antoninbas, line 55: the comment is still not correct. The `length` parameter is required, and there is no special value that will use the "minimum required number of bytes".

The Pyhton documentation has an exmple of how to achieve this however. It would look like:
```python
n.to_bytes(length = (n.bit_length() + 7) // 8, byteorder='big')
```

proto/ptf/base_test.py
```diff
@@ -49,35 +50,30 @@ def __get__(self, instance, owner):
...
     parameter.  Return a string with binary contents expected by the
     Python P4Runtime client operations.  If 'n' does not fit in
     'length' bytes, it is represented in the fewest number of bytes it
     does fit into without loss of precision.  It always returns a
     string at least one byte long, even if value=width=0."""
-    h = '%x' % n
-    s = ('0'*(len(h) % 2) + h).zfill(length*2).decode('hex')
-    return s
+    if length is None:
+        length = (x.bit_length() + 7) // 8
```
antoninbas, line 62: I think you meant `n.bit_length()` here

proto/ptf/base_test.py
```diff
@@ -49,35 +50,30 @@ def __get__(self, instance, owner):
...
+    return int.to_bytes(n, length=length, byteorder='big')
```
antoninbas, line 63: is there a difference between this and `n.to_bytes`?

proto/ptf/base_test.py
```diff
@@ -49,35 +50,30 @@ def __get__(self, instance, owner):
...
-# integers, this should only require removing the zfill() call below.
-def stringify(n, length):
+def stringify(n, length=None):
     """Take a non-negative integer 'n' as the first parameter, and a
     non-negative integer 'length' in units of _bytes_ as the second
     parameter.  Return a string with binary contents expected by the
     Python P4Runtime client operations.  If 'n' does not fit in
     'length' bytes, it is represented in the fewest number of bytes it
     does fit into without loss of precision.  It always returns a
     string at least one byte long, even if value=width=0."""
```
antoninbas, line 60: can you update this function comment to indicate what it means when `length=None`?

proto/ptf/base_test.py
rst0git: No, the result is the same:
```Python
>>> n = 3
>>> n.to_bytes()
b'\x03'
n.to_bytes(length=2)
b'\x00\x03'
>>> int.to_bytes(3)
b'\x03'
>>> int.to_bytes(3, length=2)
b'\x00\x03'
```

antoninbas (approved):

jafingerhut:
Was a file with the same name `base_test.py` updated in p4lang/p4c, perhaps a year or two ago?

rst0git:
> Was a file with the same name `base_test.py` updated in p4lang/p4c, perhaps a year or two ago? If so, perhaps keeping the identical would be advisable?

@jafingerhut Thank you for pointing this out! I've updated the pull request to keep `stringify()` identical.

PR #602, opened 2023-08-18
Fix build for Protobuf versions 3.12 through 3.14

antoninbas:
P4Runtime Protobufs now require support for proto3 optional fields, which require the `--experimental_allow_proto3_optional` protoc flag for versions 3.12 through 3.14. The flag is no longer required starting with Protobuf version 3.15.

By default, Ubuntu 22.04 comes with Protobuf 3.12.

We also ensure that the installed Protobuf versions is at least 3.12, as older versions are no longer supported by P4Runtime.

### DIFF
--- a/proto/Makefile.am
+++ b/proto/Makefile.am
@@ -9,7 +9,7 @@ endif
 
 SUBDIRS = . third_party p4info frontend server $(MAYBE_PROTO_DEMO) tests
 
-PROTOFLAGS = -I$(abs_srcdir) -I$(abs_builddir)
+PROTOFLAGS += -I$(abs_srcdir) -I$(abs_builddir)
 
 # This makefile relies on the symbolic link that we create google ->
 # googleapis/google. Without it, we run into some issues with `protoc`. An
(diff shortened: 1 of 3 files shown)


antoninbas:
@rst0git do you want to take a look at this? I think it was triggered by the submodule update.
I am not sure why you didn't observe this issue when building the Debian package for Ubuntu 22.04, unless you use a more recent Protobuf version than the default one.

rst0git:
> I am not sure why you didn't observe this issue when building the Debian package for Ubuntu 22.04, unless you use a more recent Protobuf version than the default one.

I was able to fix this problem by removing [`optional`](https://github.com/rst0git/p4lang-packages/blob/52e7f2311ecccb40d2db8bfa3140379dfbb06380/p4lang-pi/rules#L32) during the build process.

rst0git (commented):
LGTM

antoninbas:
> As a workaround for this problem, I've updated the package to remove [optional](https://github.com/rst0git/p4lang-packages/blob/52e7f2311ecccb40d2db8bfa3140379dfbb06380/p4lang-pi/rules#L32) during the build process.

The issue with that is it changes the semantics of the protobuf (it will change the generated code).

PR #614, opened 2024-08-25
Upadate the README with instructions for installing  gRPC

Dscano:
I recently attempted to install PI and found that the gRPC installation instructions in the README were outdated. gRPC can now only be installed via Bazel or CMake. The set of commands provided [here](https://github.com/grpc/grpc/blob/v1.43.2/test/distrib/cpp/run_distrib_test_cmake_module_install.sh#L29C3-L29C36) allowed me to successfully install gRPC and PI with the `./configure --with-proto`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,9 +76,16 @@ make
 git clone --depth=1 -b v1.43.2 https://github.com/google/grpc.git
 cd grpc/
 git submodule update --init --recursive
-make
-[sudo] make install
-[sudo] ldconfig
+mkdir -p "cmake/build"
+pushd "cmake/build"
+cmake \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DgRPC_INSTALL=ON \
+  -DgRPC_BUILD_TESTS=OFF \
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
I have not tried out all of the cmake options for gRPC that exist in this PR, but have often used cmake to build gRPC from source code, and except for the particular cmake options shown here, these changes look close to what I have used successfully.

README.md
```diff
@@ -76,9 +76,16 @@ make
...
-[sudo] ldconfig
+mkdir -p "cmake/build"
+pushd "cmake/build"
+cmake \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DgRPC_INSTALL=ON \
+  -DgRPC_BUILD_TESTS=OFF \
+  -DgRPC_SSL_PROVIDER=package \
+  ../..
+make -j4 install
```
antoninbas, line 87: I would guess that typically you will need to `sudo` to be able to install in a system location, so maybe decouple the build and install steps to be consistent with other instructions?

antoninbas (approved):

PR #615, opened 2024-10-17
update p4runtime

grg:


antoninbas:
The bazel build has been broken on main, so that's independent of this PR.
However, it probably makes sense to wait for the P4Runtime v1.4.1 tag to be published, which should be imminent, and https://github.com/p4lang/PI/blob/main/bazel/deps.bzl should ideally be updated at the same time as the submodule. cc @smolkaj

grg:
> The bazel build has been broken on main, so that's independent of this PR. However, it probably makes sense to wait for the P4Runtime v1.4.1 tag to be published, which should be imminent, and https://github.com/p4lang/PI/blob/main/bazel/deps.bzl should ideally be updated at the same time as the submodule. cc @smolkaj

Hi @antoninbas, I hope you're doing well!

The p4lang/p4c@ec4eb5ef commit introduces `initial_default_action` to Table messages. Without updating PI to this change, the Tofino PTF test fail with this error:
```
google.protobuf.text_format.ParseError: 41:3 : Message type "p4.config.v1.Table" has no field named "initial_default_action".
```

I'll make a local Intel fork to allow us to proceed until an update here with the official 1.4.1 P4Runtime. I'm not sure if this will impact other downstream users of p4c + PI.

fruffy:
@grg I deleted the outdated cache. Looks like things are working again.

fruffy (approved):

grg:
@fruffy / @antoninbas Should I proceed with the current PR or wait for the official 1.4.1 tag on p4runtime?

fruffy:
Personally, I think it is fine to merge. It might take a while till 1.4.1 is in and I do not think PI is tied to a particular release version.

antoninbas:
I am fine with merging, but bazel.deps should be updated as part of this PR, otherwise the autotools build and the Bazel build use a different version of p4runtime.

grg:
@fruffy / @antoninbas I've updated the Bazel config. Let me know if there's anything lese I've missed.

Also, I'm not authorized to do the merge.

PR #618, opened 2025-02-02
Add SPDX license id lines to source files

jafingerhut:


jafingerhut:
Putting this in Draft status right now, as there are still a few more source files I have not added license id lines to yet.

jafingerhut:
I know nothing about Bazel except that it is a build tool, or something in that neighborhood.

Does anyone have any thoughts on why adding some comment lines to some source files would cause the Bazel build to fail, when it doesn't fail if all I do is make a small comment change in one of the BUILD files, as shown in this PR created only for the purpose of seeing if the current files, or something closer to them, passes the Bazel CI build check?  https://github.com/p4lang/PI/pull/619

antoninbas:
@jafingerhut The Bazel CI job uses a [cache](https://github.com/p4lang/PI/blob/c3abb96d61f91e4f641aecd7ba2e784870c35290/.github/workflows/bazel-build.yml#L21), which is persisted across jobs.

My guess is that the build issue you are observing for this PR is a legitimate issue, which got exposed by touching all source files and invalidating the cache. In contrast, your test PR (#619) does not exhibit the issue because if you look at the logs for the job, you will see that the job took under a minute and simply got its outputs from the Bazel cache.

I suspect that if you make a single line change to one of the files listed in the Bazel error for this PR (e.g., `lib/vector.c`), you may also get a build error.
The errors look legitimate to me, and they may not be too hard to fix if it's just some missing Bazel dependencies.

jafingerhut:
@smolkaj Do you know if anyone at Google is using the code in this repository with Bazel for building it?  I am about 1% knowledgeable of Bazel, and from Antonin's comment above, it appears like the Bazel CI build in this repo might not be configured properly via its BUILD files right now.  I am far from the best person to try to fix it.

smolkaj:
Catching up after returning from parental leave today...

We seem to currently use our own, separate BUILD rules at Google internally.

jafingerhut:
By pinning the Bazel CI run to Ubuntu 22.04, it passes again.   See this comment: https://github.com/p4lang/PI/pull/620#issue-2830143121

jafingerhut:
I'm removing draft status on this PR.  It has the one-line CI change so that the Bazel build CI test runs on Ubuntu 22.04, which seems like a reasonable change to make until/unless someone updates the Bazel build stuff to work on Ubuntu 24.04.

jafingerhut:
@antoninbas Sorry to trouble you for such a repetitive PR. 95% of it is adding one line with a comment containing "SPDX-License-Identifier: Apache-2.0" to a file that already has a copyright notice and the couple of paragraphs saying it has an Apache 2.0 license.

4.9% of it is adding that line plus a copyright notice to files that had not copyright notice, mainly a few Bash scripts.  I checked the commit logs of those and you were the only committer on all of them I can remember, going back to 2016-2017 or so.

0.1% of it is changing the CI so it only uses ubuntu-22.04, rather than ubuntu-latest, which changed from 22.04 to 24.04 a couple of months ago and I'm pretty sure is the reason the Bazel build was failing.  A similar thing happened in the p4runtime repo a couple of months back, and we used the same temporary fix until the Bazel build files were updated to work with Ubuntu 24.04, which I hope happens to this repo soon by someone who knows more about Bazel than I do.

jafingerhut:
FYI, I am working on a simple Python program that we can use later in CI to verify that all source files have an SPDX-License-Identifier line, but that will be later after we get more p4lang repos "clean" according to it: https://github.com/jafingerhut/test-spdx-checker

.github/workflows/bazel-build.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
antoninbas, line 13: haven't you already taken care of this in https://github.com/p4lang/PI/pull/620?
could you rebase this PR?

antoninbas (approved):

.github/workflows/bazel-build.yml
jafingerhut: Yes, I went ahead and threw it in this PR as well, to see if CI passed or not.  I will rebase.

jafingerhut:
Got Antonin's approval, rebased, all CI tests pass as expected.  I will go ahead and hit the merge button on this one.

PR #621, opened 2025-02-10
Add SPDX license id lines to those .m4 files that mention a license

jafingerhut:
This does not change any licenses of any files -- it only adds their SPDX license id names to files that already have them.


jafingerhut:
For comparison if you are curious, here is the SPDX description of the FSF All Permissive (FSFAP) license: https://spdx.org/licenses/FSFAP.html

antoninbas (commented):
@jafingerhut for these .m4 files, I would actually recommending not modifying them at all. We copy them verbatim from gnu.org, and I personally don't think it makes sense to add the `SPDX-License-Identifier` to them ourselves.

jafingerhut:
Closing this, as my leading contender for adding copyright and license info for files that cannot support them directly in comments (e.g. JSON data files) is the REUSE project, which stores the copyright and license info for such files in a separate project-wide file.

PR #623, opened 2025-05-16
Correct a bug with parsing PacketIn controller metadata

jafingerhut:
The bug manifests itself if bit_offset becomes 8 or larger, which happens in some cases if there are multiple fields in the packet-in header that are not a multiple of 8 bits wide.

### DIFF
--- a/proto/frontend/src/packet_io_mgr.cpp
+++ b/proto/frontend/src/packet_io_mgr.cpp
@@ -182,6 +182,10 @@ class PacketInMutate {
       generic_extract(pkt, bit_offset, bitwidth, buffer.data());
       bit_offset += (bitwidth % 8);
       pkt += (bitwidth / 8);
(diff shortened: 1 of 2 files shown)


antoninbas (approved):
LGTM, even though I'd suggest adding a corresponding test case in https://github.com/p4lang/PI/blob/main/proto/tests/test_proto_fe_packet_io.cpp

jafingerhut:
> LGTM, even though I'd suggest adding a corresponding test case in https://github.com/p4lang/PI/blob/main/proto/tests/test_proto_fe_packet_io.cpp

I made a small change to one of the existing unit tests, and verified that without the proposed fix the test fails, but with the proposed fix the test passes.

jafingerhut:
Merging this, after modifying an existing test case so that it would have caught the bug.

PR #631, opened 2026-04-17
Automated releases

chreekat:
This is modeled after the machinery in behavioral-model and ptf, with some internal cleanups.

Tbh I'm not sure which repos it's appropriate to have automatic releases for? Is PI appropriate? :) Are there more I should do?


chreekat:
CI failure is an unexpected checksum when pulling some dependency:

```
WARNING: Download from https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException Checksum was 1c162b579a423fa6876c6c5bc16d39ab4bc05e28898977a0a6af345f523f6357 but wanted 94ced8b72956591c4775ae2207a9763d3600b30d9d7446562c552f0a14a63be7
```

chreekat:
I discovered that if I start relying on a VERSION file, we'd have two sources of truth. E.g. bmv2 uses a script to read that file inside configure.ac. I think I'll just copy what bmv2 does there. Still need to check if there are more such cases of duplication.

fruffy:
>I discovered that if I start relying on a VERSION file, we'd have two sources of truth

Can you elaborate? At least for BMv2 we should get rid of autotools soon.

fruffy:
Could you also add a form of this? https://github.com/p4lang/behavioral-model/blob/main/.github/release.yml

The automated release notes are useful

chreekat:
Whoa, I didn't know about .github/release.yml. I'll add it.

I was uncertain about configure.ac. I just noticed, mechanically, that it does refer to a version number. Seems safer to keep it up to date, even if autotools is going to get the boot. Unless you think I should just remove the file already?

fruffy:
> Whoa, I didn't know about .github/release.yml. I'll add it.
> 
> I was uncertain about configure.ac. I just noticed, mechanically, that it does refer to a version number. Seems safer to keep it up to date, even if autotools is going to get the boot. Unless you think I should just remove the file already?

Sounds good to me

chreekat:
I decided to simply not touch configure.ac. Let's remove all the autoconf stuff in a dedicated PR.

For release.yml, I copied the one over from behavioral-model, then removed stuff that didn't look relevant to PI.

NB: The labels don't exist yet in this repo!

fruffy (approved):

PR #637, opened 2026-04-25
Add file-based automated PR labeling

chreekat:
Let me know if this seems useful. I can do it to the other repos, too.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,35 @@
+autoconf:
+  - changed-files:
+      - any-glob-to-any-file:
+          - configure.ac
+          - autogen.sh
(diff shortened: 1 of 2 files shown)


chreekat:
Ah, using `.github/**` for github_actions is sketchy, it overlaps with the rule for dependabot. Hm I guess that's a gray area. But I think it's good to keep them separate. I'll change the rule for github_actions to look more specifically at .github/{actions,workflows}.

chreekat:
Ok, I'm happy with this now.

chreekat:
I went ahead and added the missing labels to the project.

fruffy (approved):

PR #641, opened 2026-05-08
Use C++17 and prioritize standard mutex utilities.

c8ef:
Part of https://github.com/p4lang/behavioral-model/issues/1360.


c8ef:
cc @matthewtlam @jafingerhut for review.

c8ef:
Reason for this patch:

1. Since bmv2 already uses C++17, it seems entirely reasonable for us to adopt C++17 as well in 2026. This patch includes updates for both Bazel and Autotools.
2. With C++17, we can use the standard library mutex instead of relying on Abseil or Boost.
3. While this patch is large, it is straightforward. The majority of changes relate to m4 macros, and the remaining actual changes are easy to review.

jafingerhut:
I checked a couple of the new .m4 files you add in this PR.  I saw copyright notices in them.  Good.  I saw a brief paragraph with a license, but it was not clear whether that license was equivalent to one of the well known licenses like MIT, BSD-2-Clause, BSD-3-Clause, etc.

Can you please explain where you copied these files from?  And if the license of a file is the same as one of the licenses listed in this page, please identify which one it is: https://spdx.org/licenses/

You can copy and paste the text of a license into the text box on this page for help in identifying if the license is the same as one of those: https://tools.spdx.org/app/check_license/

Ideally it would be best if all new files can include a comment line like the following (with the correct license identifier string -- BSD-2-Clause is just an example license name on the SPDX list), independently for each new file, as they might not all have the same license:
```
# SPDX-License-Identifier: BSD-2-Clause
```

c8ef:
> I checked a couple of the new .m4 files you add in this PR. I saw copyright notices in them. Good. I saw a brief paragraph with a license, but it was not clear whether that license was equivalent to one of the well known licenses like MIT, BSD-2-Clause, BSD-3-Clause, etc.
> 
> Can you please explain where you copied these files from? And if the license of a file is the same as one of the licenses listed in this page, please identify which one it is: https://spdx.org/licenses/
> 

The source is the same as before, from the https://github.com/autoconf-archive/autoconf-archive. So, IIUC, this does not introduce any new licensing concerns.

> You can copy and paste the text of a license into the text box on this page for help in identifying if the license is the same as one of those: https://tools.spdx.org/app/check_license/
> 
> Ideally it would be best if all new files can include a comment line like the following (with the correct license identifier string -- BSD-2-Clause is just an example license name on the SPDX list), independently for each new file, as they might not all have the same license:
> 
> ```
> # SPDX-License-Identifier: BSD-2-Clause
> ```

Done.

jafingerhut:
The top level README of https://github.com/autoconf-archive/autoconf-archive says that the collection is licensed under GPL v3 or any later version.

If that is true of the files that you are adding to the PI repository, note that _if_ these were C++ source files compiled into the PI code (I know that they are not), then GPL v3 source cannot be combined with Apache-2.0, as those licenses are judged by people as likely to be incompatible if ever challenged in a court case.

Some of the GNU automake .m4 files are licensed GPL v3, but have an explicit exception that allows them to be used to build code for projects with non-GPL licenses, e.g. this one: https://github.com/p4lang/behavioral-model/blob/main/m4/ax_prefix_config_h.m4#L104-L115

Sorry to be such a stickler, but I have been involved in adding explicit copyright and license notices to all files in the repositories, and we must be VERY CAUTIOUS when adding files with a GPL license.  We need to ensure that if they are added, either they have an explicit exception that allows them to be used in a non-copylefted project, or they are used only for testing or other purposes.

I agree that probably the files you are adding here are likely to turn out to be legally safe to add, but we need to be sure.

jafingerhut:
Also, at least one of the files you are adding does not exist in the repository  https://github.com/autoconf-archive/autoconf-archive, so looking at that repository does not help one determine the license of files that aren't in there.

c8ef:
> Also, at least one of the files you are adding does not exist in the repository https://github.com/autoconf-archive/autoconf-archive, so looking at that repository does not help one determine the license of files that aren't in there.

Can you elaborate on this a bit more? I added four identical files copied directly from https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_cxx_compile_stdcxx.m4, along with the license identifier you asked me to add.

jafingerhut:
> > Also, at least one of the files you are adding does not exist in the repository https://github.com/autoconf-archive/autoconf-archive, so looking at that repository does not help one determine the license of files that aren't in there.
> 
> Can you elaborate on this a bit more? I added four identical files copied directly from https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_cxx_compile_stdcxx.m4, along with the license identifier you asked me to add.

My apologies, I think I was checking for the existence of a file in that repo that was one of the files you were removing, not one of the files you were adding.  My mistake.

jafingerhut:
>  along with the license identifier you asked me to add.

I see that you added this line to the several copies of the file you added:
```
SPDX-License-Identifier: GPL-3.0-or-later WITH Autoconf-exception-3.0
```
Now here is the really important question: What evidence do you have that that license applies to that source file?  I see nothing in the comments of that file itself that indicate it is released under that license.

c8ef:
> > along with the license identifier you asked me to add.
> 
> I see that you added this line to the several copies of the file you added:
> 
> ```
> SPDX-License-Identifier: GPL-3.0-or-later WITH Autoconf-exception-3.0
> ```
> 
> Now here is the really important question: What evidence do you have that that license applies to that source file? I see nothing in the comments of that file itself that indicate it is released under that license.

Genuinely, I don’t know. The top-level repository license suggests `GPL-3.0-or-later WITH Autoconf-exception-3.0`, but I’m not sure how confidently that can be applied to individual files that do not contain their own license headers.

What would you suggest here? Do you think we can use the newer version or not? And if so, what would be the most appropriate license identifier to use?

jafingerhut:
If you can point out what text in the rest of the https://github.com/autoconf-archive/autoconf-archive repository suggest that all files it contains are licensed under GPL-3.0-or-later WITH Autoconf-exception-3.0, I would like to take a look at that.  I am not trying to be 100% certain here -- 95% confidence and recording of what evidence we used to base our conclusion on is enough for me.

jafingerhut:
> https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_cxx_compile_stdcxx.m4

OK.  Some more data: Here are the lines specifically mentioning copying restrictions and permissions in the one source file linked above:

https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_cxx_compile_stdcxx.m4#L42-L45

Here is a copy of that text:
```
#   Copying and distribution of this file, with or without modification, are
#   permitted in any medium without royalty provided the copyright notice
#   and this notice are preserved.  This file is offered as-is, without any
#   warranty.
```
If you go to the SPDX license check page here: https://tools.spdx.org/app/check_license/
and copy the text above into it, it tells you "Perfect match! The license matches with the license ID(s): FSFAP"

The FSFAP license applies to this file: described more fully here: https://spdx.org/licenses/FSFAP.html

As far as I can tell from searches on Internet, it is perfectly fine to include a file with FSFAP license in a project that is released under Apache-2.0.

Please add this line to every copy of that file you are adding, and we should be good to go license-wise:
```
SPDX-License-Identifier: FSFAP
```
(plus any characters required to make this line a comment)

fruffy:
I am a little surprised that autoconf would cause licenses issues like this... 

Hope we can switch to 100% CMake to avoid this confusion.

jafingerhut:
autoconf isn't causing license issues -- I'm just being hyper-vigilant that we know the license of new added files, and the submitter did not know what the license was.

c8ef:
> > https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_cxx_compile_stdcxx.m4
> 
> OK. Some more data: Here are the lines specifically mentioning copying restrictions and permissions in the one source file linked above:
> 
> https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_cxx_compile_stdcxx.m4#L42-L45
> 
> Here is a copy of that text:
> 
> ```
> #   Copying and distribution of this file, with or without modification, are
> #   permitted in any medium without royalty provided the copyright notice
> #   and this notice are preserved.  This file is offered as-is, without any
> #   warranty.
> ```
> 
> If you go to the SPDX license check page here: https://tools.spdx.org/app/check_license/ and copy the text above into it, it tells you "Perfect match! The license matches with the license ID(s): FSFAP"
> 
> The FSFAP license applies to this file: described more fully here: https://spdx.org/licenses/FSFAP.html
> 
> As far as I can tell from searches on Internet, it is perfectly fine to include a file with FSFAP license in a project that is released under Apache-2.0.
> 

Thanks for your detailed explanation! I really learned a lot about open source licensing from this discussion. I hadn’t realized that the top-level license may not apply to every source file, or that the SPDX website can help identify the correct license. One minor nit: when pasting into the website, we need to remove the leading `#` characters.

> Please add this line to every copy of that file you are adding, and we should be good to go license-wise:
> 
> ```
> SPDX-License-Identifier: FSFAP
> ```
> 
> (plus any characters required to make this line a comment)

Done.

jafingerhut (approved):
LGTM from software license perspective, and tests appear to be passsing.  I leave C++ reviews up to others, if they wish to examine those details.

proto/configure.ac
```diff
@@ -58,7 +58,6 @@ AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
 
 AX_BOOST_BASE([1.54.0], [], [AC_MSG_ERROR(
               [Please install boost >= 1.54.0 (thread and system)])])
-AX_BOOST_THREAD
 AX_BOOST_SYSTEM
 AC_SUBST([BOOST_THREAD_LIB])
```
fruffy: Still needed?

fruffy (approved):

proto/configure.ac
c8ef: Good catch! Removed all occurrences.

PR #643, opened 2026-05-12
Extract common release workflows

chreekat:
p4c, behavioral-model, ptf, and PI all have identical automated release machinery. I've centralized the code into a single repo they can all share. PI is the pilot customer. The playground for common code is https://github.com/chreekat/auto-release-playground .

Future work will be to move the central repo out of my personal namespace. Or maybe we reuse the p4lang/.github for the purpose.


chreekat:
Sadly, we can't actually test the change until it lands on main. A real annoying limitation of GitHub.

Once it *does* land, we can easily test part of it by running a manual workflow to create a PR that bumps the version. We don't have to actually accept the PR.

fruffy:
Yes, we can use https://github.com/p4lang/.github for this.

fruffy (approved):

PR #646, opened 2026-05-14
Migrate from `boost::optional` to `std::optional`

c8ef:
This is a prerequisite for refreshing the bazel build of PI, which is a prerequisite for the bazel build of bmv2.


c8ef:
cc @fruffy @jafingerhut @matthewtlam for review.

fruffy (approved):

PR #647, opened 2026-05-16
Refresh bazel build of PI.

c8ef:
This patch mainly does two things:

- Updates the Bazel build for the new Bazel version (targeting 7 to 9)
- Formats the Bazel files with Buildifier

Closes #629.


c8ef:
cc @jafingerhut @fruffy for review. I noticed the Bazel CI has been manually disabled - could you help re-enable it so the CI changes in this PR can be verified?

cc @matthewtlam @smolkaj as potential users of the Bazel build.

fruffy:
> cc @jafingerhut @fruffy for review. I noticed the Bazel CI has been manually disabled - could you help re-enable it so the CI changes in this PR can be verified?

Done. Might require a dummy push to retrigger.

c8ef:
> > cc @jafingerhut @fruffy for review. I noticed the Bazel CI has been manually disabled - could you help re-enable it so the CI changes in this PR can be verified?
> 
> Done. Might require a dummy push to retrigger.

Thanks for the quick response!

c8ef:
The bazel CI works now!

proto/third_party/BUILD
```diff
@@ -1,21 +1,25 @@
...
-            "fmt/ostream.h",
-            "fmt/posix.cc",
-            "fmt/posix.h",
-            "fmt/printf.cc",
-            "fmt/printf.h",
-            "fmt/string.h",
-            "fmt/time.h",
-            "fmt/_assert.h",
-            "fmt/_assert.cc"],
+    srcs = [
```
fruffy, line 9: Bazel does have fmt available, maybe we can replace this old dependency in the future: https://registry.bazel.build/modules/fmt

fruffy (approved):

proto/third_party/BUILD
c8ef: Yes, but we still need to support autotools... I will follow up on this later.

PR #654, opened 2026-06-20
Fix null pointer dereference in default idle-timeout notification

Devansh-567:
## Description
This PR fixes a critical null pointer dereference bug in `pi_table_idle_timeout_notify` that causes a deterministic segmentation fault whenever only a default idle-timeout callback is registered.

### Problem
When a device-specific callback is not found, `cb_data` evaluates to `NULL`. The code appropriately falls back to checking for a global/default callback (`default_cb_data`). However, during the invocation of this default callback, the function mistakenly attempts to pass `cb_data->cookie` instead of `default_cb_data->cookie`. Because `cb_data` is known to be `NULL` at this point, it triggers a null pointer dereference.

### Solution
Updated the default callback invocation on line 471 to correctly reference `default_cb_data->cookie` instead of `cb_data->cookie`.

---

## Type of Change
- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/pi_tables.c
+++ b/src/pi_tables.c
@@ -467,8 +467,8 @@ pi_status_t pi_table_idle_timeout_notify(pi_dev_id_t dev_id,
   }
   const cb_data_t *default_cb_data = cb_mgr_get_default(&cb_mgr);
   if (default_cb_data->cb) {
-    ((PIIdleTimeoutCb)(default_cb_data->cb))(dev_id, table_id, match_key,
-                                             entry_handle, cb_data->cookie);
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
LGTM.  From manual inspection of the code, I agree that it looks like this code must have been attempting to do a null pointer dereference before, if it was ever reached.

jafingerhut:
There is some kind of style check failure occurring in CI.  It is not obvious to me from the logs exactly why this is.  Are you able to try running the checking Bash script that is failing in CI locally, and see if it is easier to determine in that way more details, sufficient to determine what might need to change in order to pass that check?

PR #655, opened 2026-06-20
Fix vector remove element index

Devansh-567:
## Description
This PR addresses a severe bug in `vector_remove_e` where a byte offset is incorrectly passed as an element index to `vector_remove`, breaking element removal routines.

### Problem
In `lib/vector.c`, the `vector_remove_e` function calculates the position of an element pointer by subtracting `v->data` from `e`. This calculates a raw **byte offset**, not an element index. Passing this byte offset directly to `vector_remove` causes out-of-bounds asserts or memory corruption for any index greater than 0. This directly impacts core system behaviors such as device removal sequences (`src/pi.c:266`).

### Solution
- Divided the computed byte distance by `v->e_size` to calculate the actual item index.
- Added an explicit alignment assertion (`assert(offset % v->e_size == 0)`) to ensure the pointer target maps properly onto a clean element boundary.

---

## Type of Change
- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/lib/vector.c
+++ b/lib/vector.c
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *   http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -94,7 +94,9 @@ void vector_remove(vector_t *v, size_t index) {
 
 void vector_remove_e(vector_t *v, void *e) {
   assert(e >= v->data);
-  size_t index = (char *)e - (char *)v->data;
+  size_t offset = (char *)e - (char *)v->data;
+  assert(offset % v->e_size == 0);
+  size_t index = offset / v->e_size;
   vector_remove(v, index);
 }
 


jafingerhut:
It is best to create PRs with independent changes.  This PR includes your change from another PR you recently created.  I suspect you are not using git correctly, e.g. you are probably creating a new branch for one fix, committing and pushing it, creating a PR, and then for the second fix you are creating a new branch ON TOP OF the branch you created for the first fix.  Instead you should do `git checkout main`, THEN create a second new branch with a different name than the one you created for the first fix.

Devansh-567:
> It is best to create PRs with independent changes. This PR includes your change from another PR you recently created. I suspect you are not using git correctly, e.g. you are probably creating a new branch for one fix, committing and pushing it, creating a PR, and then for the second fix you are creating a new branch ON TOP OF the branch you created for the first fix. Instead you should do `git checkout main`, THEN create a second new branch with a different name than the one you created for the first fix.

Hi @jafingerhut,

Thank you so much for the guidance! You caught exactly what I did wrong. I accidentally branched the second fix on top of the first one instead of starting fresh from a clean `main` branch. 

I have fixed my local history by checking out a clean `main`, cherry-picking only the vector fix commit, and force-pushing it here. This PR now completely isolates the `vector_remove_e` fix as intended. 

I really appreciate you taking the time to explain the proper workflow, definitely a great learning moment for me!

jafingerhut:
Thanks for your interest in looking for and fixing problems in the PI repository code.

For all of the PRs you have raised recently, how did you discover the problem?

Do you have any test cases that were failing without the fix, but passing with the fix?  Such a test need not be in the PI repository by itself -- for example, it might be that when compiling behavioral-model with PI and running some test using that combination, there is a failure when performing some appropriate control plane operation that triggers the problem.

It is not _required_ to have a test case that fails without the fix, and passes with the fix, but it does help provide additional context and justification for the change that a reviewer can use to guide their judgement.  If you have no such test case, it is OK to simply say so, but then knowing how you discovered the problem is also helpful to know.

Devansh-567:
> Thanks for your interest in looking for and fixing problems in the PI repository code.
> 
> For all of the PRs you have raised recently, how did you discover the problem?
> 
> Do you have any test cases that were failing without the fix, but passing with the fix? Such a test need not be in the PI repository by itself -- for example, it might be that when compiling behavioral-model with PI and running some test using that combination, there is a failure when performing some appropriate control plane operation that triggers the problem.
> 
> It is not _required_ to have a test case that fails without the fix, and passes with the fix, but it does help provide additional context and justification for the change that a reviewer can use to guide their judgement. If you have no such test case, it is OK to simply say so, but then knowing how you discovered the problem is also helpful to know.



> Thanks for your interest in looking for and fixing problems in the PI repository code.
> 
> For all of the PRs you have raised recently, how did you discover the problem?
> 
> Do you have any test cases that were failing without the fix, but passing with the fix? Such a test need not be in the PI repository by itself -- for example, it might be that when compiling behavioral-model with PI and running some test using that combination, there is a failure when performing some appropriate control plane operation that triggers the problem.
> 
> It is not _required_ to have a test case that fails without the fix, and passes with the fix, but it does help provide additional context and justification for the change that a reviewer can use to guide their judgement. If you have no such test case, it is OK to simply say so, but then knowing how you discovered the problem is also helpful to know.

Hi @jafingerhut,

Thanks for the encouraging words! 

I discovered these issues while doing a code review and structural audit of the library's memory management and callback routines. I was tracking how the control plane handles device tracking and callback lifecycles, which led me to inspect `src/pi_tables.c` and `lib/vector.c` closely.

For the specific bugs:
* **PR #654 (Null Dereference):** This stood out during a static analysis sweep. Because the block explicitly executes only when `cb_data` evaluates to `NULL` to fall back to `default_cb_data`, seeing `cb_data->cookie` invoked inside that exact condition immediately flagged a logical contradiction and an inevitable segmentation fault.
* **PR #655 (Vector Index):** While reviewing how `vector_remove_e` processes elements, I noticed that the raw pointer subtraction `(e - v->data)` evaluates to a raw byte distance in standard memory space rather than an array index offset. Passing a raw byte value directly into a function expecting an element index is a structural mismatch that inevitably corrupts adjacent memory blocks or triggers bounds asserts for any item beyond index 0.

I don't have a standalone, isolated test case framework configured locally for the behavioral-model integration, so these fixes were derived purely from source-code tracing and structural validation. I wanted to patch these underlying stability vulnerabilities proactively before they disrupted runtime operations.

jafingerhut (approved):
Looks correct to me, and does look incorrect in the current code before this change.

PR #657, opened 2026-06-24
 fix(lib): resolve read_file safety bugs, ensure cross-platform compatibility, and add unit tests

Devansh-567:
## Description 
This PR refactors `lib/read_file.c` to handle file reading with robust, cross-platform safety guarantees. It flattens the error handling architecture into a clean, idiomatic allocation-cleanup structure and introduces a comprehensive regression test suite using the Unity framework to prevent future regressions.

### The Problem
1. **Cross-Platform Failure:** Opening files with `"r"` causes subtle bugs on Windows due to line-ending translations, resulting in corrupted string buffers or trailing garbage memory.
2. **Silent Failure Paths:** If `ftell` returned an error (`-1`) or `fseek` failed, the function skipped allocation but still executed downstream logic blindly or exited silently without setting an explicit state.
3. **Missing Bound Validation:** `fread` was trusted implicitly without verifying if the returned read size matched the expected buffer size, risking unhandled read/hardware stream faults.
4. **Empty File Handling:** Empty files did not explicitly guarantee a clean, immediate NUL-termination allocation.

### The Solution
1. **Binary Mode Force:** Switched `fopen` mode to `"rb"` to ensure exact byte-for-byte read parity across Windows, macOS, and Linux.
2. **Flattened Cleanup (`goto cleanup`):** Replaced deeply nested `if` statements with a reliable single-exit cleanup cascade, ensuring `fp` is always closed and partial allocations are cleanly freed on failure paths.
3. **Strict Validation:** Added defensive checks for `fseek` bounds, negative `ftell` sizes, and `malloc` results.
4. **Robust Termination:** Guaranteed that successful reads always return a valid, NUL-terminated (`\0`) buffer, even if the target file contains 0 bytes.

---

### Modified Tracked Files
* **`lib/read_file.c`**: Core logic overhaul. Replaced the nested blocks with verified error jumps, added binary read modes, and implemented strict allocation checking.
* **`tests/Makefile.am`**: Updated Autotools configurations to compile `test_read_file`, define its preprocessor macros (`-DTEST_READ_FILE`), and add it to the automated execution matrix (`TESTS`, `check_PROGRAMS`).
* **`tests/main.c`**: Externally declared `test_read_file()` and conditionally wired it into the centralized test runner environment under `#ifdef TEST_READ_FILE`.

### Newly Created Files
* **`tests/test_read_file.c`**: A brand new dedicated Unity testing file executing regression checks against missing files, empty inputs, and standard text sequences.

---

## How I Found This Problem
While reviewing the core utilities in `lib/`, I noticed that `read_file` opened files in text mode (`"r"`). On Windows environments, this triggers automatic CRLF (`\r\n`) to LF (`\n`) translation, causing `ftell` byte sizes to mismatch the actual bytes read by `fread`. Intrigued by how the system handles edge cases like empty or malformed files under this setup, I ran a static analysis pass over the error boundaries. This revealed nested conditional paths that could fail silently, leave buffers non-NUL-terminated, or improperly manage file streams on sudden allocation failures.

---

### Manual Compilation Passes
Because this working copy does not currently have a generated top-level root Makefile available to kick off a global `make check`, verification was executed directly via localized compiler flags:

1. **Core Library Compilation Pass:**
```bash
   gcc -std=c99 -Wall -Wextra -Ilib -c lib/read_file.c
```
<i>Result: Passed with zero errors and zero warnings.</i>
<ol><li>Test Harness Compilation Pass:</li></ol>

```bash
gcc -std=c99 -Wall -Wextra -Ilib -Ithird_party/unity/include -c tests/test_read_file.c
```
<i>Result: Passed with zero errors and zero warnings.</i>

<ul>
- Scenarios Covered in test_read_file.c
<li> Non-Existent File: Asserts that trying to read a missing file returns NULL immediately. </li>
<li> Zero-Byte File: Asserts that an empty file successfully returns a valid pointer containing an immediate \0 terminator. </li>
<li> Standard ASCII Text File: Asserts that standard text reads return matching buffers with verified string content lengths. </li> </ul>

---

## Questions for Maintainers / Exceptions
Copyright Notice Year: In the newly created test file (tests/test_read_file.c), I preserved the existing repository header convention which specifies a copyright date of 2013. Should this header block be updated to reflect the current year (2026), or do you prefer keeping the legacy template block across all test additions? Let me know and I can modify it immediately!

And if any cosmetic changes needed, do let me know :)

### DIFF
--- a/tests/test_read_file.c
+++ b/tests/test_read_file.c
@@ -1,0 +1,75 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "read_file.h"
+
+#include "unity/unity_fixture.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define EMPTY_FILE "read_file_empty.tmp"
+#define TEXT_FILE "read_file_text.tmp"
+#define MISSING_FILE "read_file_missing.tmp"
+
+TEST_GROUP(ReadFile);
+
+TEST_SETUP(ReadFile) {
+  remove(EMPTY_FILE);
+  remove(TEXT_FILE);
+  remove(MISSING_FILE);
+}
+
+TEST_TEAR_DOWN(ReadFile) {
+  remove(EMPTY_FILE);
+  remove(TEXT_FILE);
+  remove(MISSING_FILE);
+}
+
+TEST(ReadFile, MissingFile) { TEST_ASSERT_NULL(read_file(MISSING_FILE)); }
+
+TEST(ReadFile, EmptyFile) {
+  FILE *fp = fopen(EMPTY_FILE, "wb");
+  TEST_ASSERT_NOT_NULL(fp);
+  TEST_ASSERT_EQUAL_INT(0, fclose(fp));
+
+  char *contents = read_file(EMPTY_FILE);
+  TEST_ASSERT_NOT_NULL(contents);
+  TEST_ASSERT_EQUAL_STRING("", contents);
+  free(contents);
+}
+
+TEST(ReadFile, TextFile) {
+  const char *expected = "line 1\nline 2\n";
+  FILE *fp = fopen(TEXT_FILE, "wb");
+  TEST_ASSERT_NOT_NULL(fp);
+  TEST_ASSERT_TRUE(fputs(expected, fp) >= 0);
+  TEST_ASSERT_EQUAL_INT(0, fclose(fp));
+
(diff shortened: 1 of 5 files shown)


tests/test_read_file.c
```diff
@@ -0,0 +1,75 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
```
jafingerhut: If you wrote a substantial portion of this file, it is reasonable to make it copyright either by you as an individual, or if you did this as a work for hire for some company, they may want to be the copyright holder.  If you can claim yourself as a copyright holder, feel free to do that, or you are welcome to choose "The P4 Language Consortium" instead.  Barefoot Networks no longer exists as a corporation -- the only reason to keep this copyright holder is if this is a completely trivial change to an existing file with this copyright notice.

jafingerhut:
This all looks reasonable to me.  I noticed that there is a call to read_file() in file src/p4info/p4info.c that does not check whether the return value is NULL.  Would you mind also adding in that check?   I am not sure, but hopefully there is some reasonable error status value that can be returned from the function that calls read_file() in there.

Devansh-567:
Hi @jafingerhut,
Thank you for the review and feedback!

- Copyright Notice: Updated the header block in `tests/test_read_file.c` to remove the legacy Barefoot Networks notice and established it under The P4 Language Consortium & Devansh Singh.

- Missing NULL Guard: Added a check for config_tmp == NULL inside `pi_add_config_from_file()` (src/p4info/p4info.c) which now safely interrupts execution and gracefully returns PI_STATUS_CONFIG_READER_ERROR instead of risking an upstream configuration fault.

Let me know if there are any other modifications or cosmetic adjustments you'd like to see!

jafingerhut (approved):
LGTM

PR #659, opened 2026-06-25
fix(bmv2_json_reader): free cJSON root and reader_state on all error paths

Devansh-567:
### Problem

`pi_bmv2_json_reader` parses a cJSON tree and initializes a `reader_state_t` at the beginning of the function, but several early `return` statements exit without releasing allocated resources.

Specifically:

1. **`pi_bmv2_json_reader`** leaks `root` when `check_json_version` fails. Additionally, if any of the seven `read_*` helper functions fail, both `root` and `state` are leaked because the function returns immediately without performing cleanup.

2. **`read_counters`** allocates two temporary cJSON arrays, `counters_` and `direct_counters_`, during its first pass. If the `is_direct` field is missing while iterating through the JSON objects, the function returns immediately, leaking any arrays that have already been allocated.

3. **`read_meters`** exhibits the same issue as `read_counters`. The temporary arrays `meters_` and `direct_meters_` are allocated locally, but an early return caused by a missing `is_direct` field bypasses cleanup and leaks the allocated arrays.

4. **`exclude_header`** contains commented-out code after an unconditional `return false`. Since this code can never be executed, it is unreachable and potentially misleading to future readers.

---

### How I Found It

While exploring the PI codebase, I noticed that `pi_bmv2_json_reader` allocates both a `cJSON *root` object and a `reader_state_t` structure near the start of the function, yet contains multiple direct `return` statements throughout its error paths.

Tracing each failure path manually showed that a failed version check leaks `root`, while failures in any of the `read_*` helper functions leak both `root` and `state`. Further inspection of `read_counters` and `read_meters` revealed a similar pattern involving locally allocated temporary cJSON arrays.

---

#### `pi_bmv2_json_reader`

The duplicated cleanup blocks were consolidated into a shared two-label cleanup structure:

```c
cleanup:
  destroy_reader_state(&state);

cleanup_root:
  cJSON_Delete(root);
  return status;
```

Failures that occur after `state` has been initialized now jump to `cleanup`, while the version-check failure jumps directly to `cleanup_root`, since `state` has not yet been initialized at that point. This ensures that all allocated resources are released exactly once on every exit path.

#### `read_counters` and `read_meters`

The duplicated cleanup logic in these functions was replaced with a shared `goto cleanup_first_pass` pattern. This guarantees that all temporary arrays are freed regardless of where an error occurs and avoids repeating identical cleanup code at multiple return sites.

#### `exclude_header`

The unreachable commented-out block following `return false` was removed.

---

### Files Changed

- `src/config_readers/bmv2_json_reader.c`

---

### Testing

- Existing unit tests pass unchanged.
- No functional behavior is modified.
---

### DIFF
--- a/LICENSES/BSD-3-Clause.txt
+++ b/LICENSES/BSD-3-Clause.txt
@@ -1,0 +1,11 @@
+Copyright (c) <year> <owner>. 
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+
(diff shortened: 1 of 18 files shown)


jafingerhut:
About 90% of the lines of change seem to be changes in whitespace.  The one failing CI test at this point looks like it is because an automated code formatting check is failing, perhaps because of these whitespace changes?

I am happy to take a look at this in more detail after all CI checks pass.  If that requires reducing the number of whitespace-only changes, that will just make the reviewing job even more quick :-)

PR #660, opened 2026-06-27
fix(core): validate all direct resource configurations during table entry validation

sanket-jadhav-cse:
### Summary

This PR fixes a validation bug in `check_direct_res_config()` where only the first direct resource configuration was validated, even when multiple direct resource configurations were provided.

It also adds a regression test to prevent this behavior from recurring.

---

### Problem

`check_direct_res_config()` iterates over all entries in `pi_direct_res_config_t`, but always reads:

```cpp
direct_res_config->configs[0].res_id
```

instead of using the current loop index.

As a result, only the first direct resource configuration is validated repeatedly, while subsequent configurations are skipped.

---

### Root Cause

The validation loop correctly iterates over `num_configs`, but the array access mistakenly uses a constant index (`0`) instead of the loop variable (`i`).

This causes invalid direct resource IDs after the first entry to bypass validation.

---

### Solution

Replace

```cpp
configs[0].res_id
```

with

```cpp
configs[i].res_id
```

so that every configured direct resource is validated.

Additionally, this PR adds a regression test covering a table entry with multiple direct resource configurations where the first resource is valid and the second resource is invalid.

The regression test verifies that validation correctly returns:

```
PI_STATUS_NOT_A_DIRECT_RES_OF_TABLE
```

instead of incorrectly accepting the request.

---

### Files Changed

* `src/pi_tables.c`

  * Fix array indexing in `check_direct_res_config()`.

* `tests/frontends/generic/test.c`

  * Add regression test covering multiple direct resource validation.

---

### Verification

The fix was verified by reproducing the issue before the change and comparing the behavior after applying the fix.

Observed behavior:

* Before the fix:

  * Only the first direct resource configuration was validated.
  * Invalid subsequent resource IDs were incorrectly accepted.

* After the fix:

  * Every direct resource configuration is validated.
  * Invalid resource IDs correctly return `PI_STATUS_NOT_A_DIRECT_RES_OF_TABLE`.

The relevant unit tests, including the added regression test, pass successfully in my local environment.

---

If there are any preferred changes to the regression test structure or placement, I'd be happy to update the PR accordingly.

### DIFF
--- a/tests/frontends/generic/test.c
+++ b/tests/frontends/generic/test.c
@@ -23,12 +23,16 @@
 #include "PI/int//serialize.h"
 #include "PI/p4info.h"
 #include "p4info/actions_int.h"
+#include "p4info/counters_int.h"
 #include "p4info/p4info_struct.h"
 #include "p4info/tables_int.h"
 
+#include "PI/int/pi_int.h"
+#include "PI/pi.h"
 #include "unity/unity_fixture.h"
 
 #include <stdlib.h>
+#include <string.h>
 
 #define DEFAULT_TABLE_SIZE 1024
 #define DEFAULT_TABLE_IS_CONST false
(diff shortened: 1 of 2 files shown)


tests/frontends/generic/test.c
```diff
@@ -295,10 +299,93 @@ TEST_GROUP_RUNNER(FrontendGeneric_Adata) {
...
+                      DEFAULT_TABLE_IDLE_TIMEOUT);
+  pi_p4info_table_add_action(p4info, tid, aid,
+                             PI_P4INFO_ACTION_SCOPE_TABLE_AND_DEFAULT);
+
+  pi_p4_id_t ctr0 = pi_make_direct_counter_id(0);
+  pi_p4_id_t ctr1 = pi_make_direct_counter_id(1);
+  pi_p4info_direct_counter_add(p4info, ctr0, "counter0",
+                               PI_P4INFO_COUNTER_UNIT_BOTH, 0, tid);
+  pi_p4info_direct_counter_add(p4info, ctr1, "counter1",
+                               PI_P4INFO_COUNTER_UNIT_BOTH, 0, tid);
```
jafingerhut: If I understand correctly, this test case is creating a table with two direct counters associated with it?

If so, I see that this is a good way to exercise the changed code.  No argument there.

I would suggest that if it also covers the code fix, it is much more common in practice for a table that has more than one direct resource to have one direct counter and one direct meter, rather than having two direct counters.  If it is easy to modify this new test to do that instead, that would be great.

jafingerhut: If that change is any significant trouble, just reply and let me know and I can approve it as it is now.

sanket-jadhav-cse: Thank you for your recommendation! It is logical indeed. I will modify the regression test to include one counter and one meter directly rather than indirectly, because this approach better resembles reality while testing the validation flow. I will commit my changes shortly.

jafingerhut (approved):
LGTM

PR #669, opened 2026-09-02
Update name of Ubuntu 24.04 Docker image used as input

jafingerhut:


.github/workflows/test.yml
```diff
@@ -53,8 +49,9 @@ jobs:
     - name: Run unit tests
       run: |
         docker run -w /PI pi bash -c \
-          'apt-get update && apt-get install -y python3-pip && \
-          python3 -m pip install ptf scapy grpcio googleapis-common-protos protobuf==3.20 && \
+          'sudo apt-get update && sudo apt-get install -y build-essential python3-dev &&
```
fruffy: sudo is not available in docker iirc.

Looks like this should be part of the third-party image, if it is missing?

fruffy: For the record, also okay to add it here, but likely should be in the build file. 

Not when we run tests. If the tests fail, our image is incomplete

jafingerhut: Thanks for the advice.  Right now I am looking into some failures to compile the PI code when `configure` is used to enable all of the optional pieces.  I have compiled PI with a subset of its options enabled for years on Ubuntu 24.04 and gotten it to pass tests when linked with simple_switch_grpc and run through all of p4c's tests, for example, but apparently there is PI C++ code that does not compile at all with the default version of GCC installed by apt on Ubuntu 24.04, that I haven't ever tried to compile before now.

We are experiencing this because either (a) no one has ever tried this before and seen the failures, or (b) they have experienced them, but haven't bothered to report them.

jafingerhut:
@fruffy This should be passing all tests now.

I also did these changes:
+ updated the version of clang tested.  The older ones did not even have apt packages available on Ubuntu 24.04
+ disabled `--with-proto` configure option in Dockerfile.bmv2.  Enabling that causes compile-time errors with some source files that I don't know to fix at this time.
+ disabled running of `./tools/check_style.sh` at the end.  There are failures there that I haven't looked into how to fix, but the better answer there is probably to use code style/lint checking tools that are used in other C++ and Python code in other p4lang projects.

.github/workflows/test.yml
```diff
@@ -34,12 +34,10 @@ jobs:
     strategy:
       matrix:
         include:
+        - cxx: clang++-18
```
fruffy: Can we make this generic like we do with gcc?

.github/workflows/test.yml
```diff
@@ -53,13 +51,14 @@ jobs:
     - name: Run unit tests
       run: |
         docker run -w /PI pi bash -c \
-          'apt-get update && apt-get install -y python3-pip && \
-          python3 -m pip install ptf scapy grpcio googleapis-common-protos protobuf==3.20 && \
+          'apt-get update && apt-get install -y build-essential python3-dev &&
```
fruffy, line 54: This should be part of the build step, not the test step. If the tests require this to pass we are shipping a broken image

Dockerfile.bmv2
```diff
@@ -66,7 +66,10 @@ WORKDIR /PI/
 RUN apt-get update && \
     apt-get install -y --no-install-recommends $PI_DEPS $PI_RUNTIME_DEPS && \
     ./autogen.sh && \
-    ./configure --enable-Werror --with-bmv2 --with-proto --with-cli --with-internal-rpc --with-sysrepo && \
+    # TODO: Change --without-proto to --with-proto when we learn how
```
fruffy, line 70: What is the actual error here? Maybe it is a trivial fix

.github/workflows/test.yml
jafingerhut: Done in commit 16, passes CI tests still.

Dockerfile.bmv2
jafingerhut: You can see the errors in the logs of this similar PR, which I created in parallel to try out some variations on this one.

jafingerhut: https://github.com/p4lang/PI/pull/670

jafingerhut: I have since found fairly small changes to the C++ code to keep the `proto` option enabled during PI conigure step.  The changes are in the PR #670, which I suspect may replace this one soon.  I plan to make a few more commits to #670 soon to try to move the installation of Python packages into the build step rather than the test step, as you commented on for this PR.

jafingerhut:
Closing this PR, as I recommend reviewing PR #670 instead.

PR #672, opened 2026-09-12
Enable CI tests to fail when test commands fail

jafingerhut:
The step named 'Run tests' in test.yml were failing, but this was masked because of the way that `make check -j2` was invoked.  This PR changes how it is invoked so that failure of that command causes the CI test to fail, too.

This exposed that with Ubuntu 24.04, some of those tests were failing.  The only reason they fail is because several of them enable the use of valgrind on some of the tests, to detect things that valgrind can detect (e.g. accessing memory that has been freed, or never been allocated, or attempts to free already-freed memory, etc.).  Those tests pass if we disable valgrind checks.  This PR disables valgrind for those tests.  If someone wants to examine the C++ code to see why valgrind is reporting those issues, I'd recommend saving that task for a later PR for someone interested in doing so.

Also update googletest from the former version used (dated 2019-May-27) to the version tagged v1.15.2 (dated 2024-Jul-31).


fruffy (approved):

chreekat (approved):
I see an issue has already been created about valgrind: https://github.com/p4lang/PI/issues/673

